PR #1, opened 2016-02-09
First real functional test journey

servingUpAces:
- GET home page
- GET sign-in
- POST sign-in
- GET two-factor
- POST two-factor

This test is using an existing account with a service to sign in.
The request need to use a session, also needs to retrieve the CSRF token from the previous GET in order to post the next response.
In order to run these test locally you will need some environment variables which I can provide.


tests/test_signin_flow.py
```diff
@@ -0,0 +1,62 @@
...
+        index_resp = client.get(base_url)
+        assert index_resp.status_code == 200
+        assert 'GOV.UK Notify' in index_resp.text
+
+        get_sign_resp = client.get(base_url + '/sign-in')
+        # print('headers: {}'.format(get_reg_resp.headers))
+        # it is possible to assert that headers are set properly here.
+
+        assert 200 == get_sign_resp.status_code
+        assert '<p>If you do not have an account, you can <a href="register">register for one now</a>.</p>' in \
```
quis: I’d rather we didn’t test for specific HTML with string matching. It’s fragile. Either:
- just test for the string
- use something like XPath (ewww) or BeautifulSoup to parse the markup and make sure the HTML you want is there.

In this specific case, I don’t think that we care that there is a `<p>` around this message. A good candidate of testing for specific HTML would be making sure a flash message was wrapped in something with `class='banner-dangerous` (for example).

minglis: Is page title testing useful here?

tests/test_utils.py
```diff
@@ -0,0 +1,11 @@
+from tests.utils import get_sms, delete_sms_messge
```
minglis: Do we need this? Or should we remove it? Is it useful?

PR #3, opened 2016-02-12
Assert page title is correct for get requests.

servingUpAces:


tests/test_signin_flow.py
```diff
@@ -15,13 +16,13 @@ def test_sign_in_journey():
...
         assert index_resp.status_code == 200
-        assert 'GOV.UK Notify' in index_resp.text
+        assert 'GOV.UK Notify' == find_page_title(index_resp.text)
 
         get_sign_resp = client.get(base_url + '/sign-in')
         # print('headers: {}'.format(get_reg_resp.headers))
         # it is possible to assert that headers are set properly here.
-
         assert 200 == get_sign_resp.status_code
+        assert 'Sign in - GOV.UK Notify'
```
quis: Think this line is missing the `== find_page_title()` bit.

PR #4, opened 2016-02-16
Registration flow

mannickutd:


tests/test_register_flow.py
```diff
@@ -0,0 +1,113 @@
...
+
+def test_register_journey():
+    '''
+    Runs through the register flow creating a new user.
+    '''
+    client = session()
+    base_url = Config.NOTIFY_ADMIN_URL
+    index_resp = client.get(base_url)
+    assert index_resp.status_code == 200
+    assert 'GOV.UK Notify' in index_resp.text
```
quis: Every page is going to have ‘GOV.UK Notify’ in it. It’s in the page `<title>`. Would be better to test for something unique to the home page, like ‘Set up an account’.

tests/test_register_flow.py
```diff
@@ -0,0 +1,113 @@
...
+def _get_email_code(email, pwd, email_folder):
+    timer = 0
+    gimap = None
+    while timer < Config.EMAIL_TIMEOUT:
+        try:
+            gimap = imaplib.IMAP4_SSL('imap.gmail.com')
+            try:
+                rv, data = gimap.login(email, pwd)
+            except imaplib.IMAP4.error:
+                    pytest.fail("Login to email account has failed.")
```
quis: Indentation not consistent.

tests/test_register_flow.py
```diff
@@ -0,0 +1,113 @@
...
+            rv, data = gimap.select(email_folder)
+            rv, data = gimap.search(None, "ALL")
+            ids_count = len(data[0].split())
+            if ids_count > 1:
+                pytest.fail("There is more than token email")
```
quis: More than _one_

tests/test_register_flow.py
mannickutd: Yes you are right, It's more of just a backup test for the 200 response. The 200 response is probably good enough.

tests/test_register_flow.py
mannickutd: Yes each test should be cleaning up after itself and for any reason if we have more than one email in that folder we want to fail.

tests/test_register_flow.py
```diff
@@ -0,0 +1,113 @@
...
+            elif ids_count == 1:
+                num = data[0].split()[0]
+                rv, data = gimap.fetch(num, '(UID BODY[TEXT])')
+                msg = email_lib.message_from_bytes(data[0][1])
+                gimap.store(num, '+FLAGS', '\\Deleted')
+                gimap.expunge()
+                return msg.get_payload().strip()
```
quis: It works, but the control flow here isn’t easy to understand. The `finally` runs _after_ the return statement, and can potentially override what the method returns. Also, if you call `pytest.fail` within the `try` block then the `finally` never gets run, and the IMAP connection is never closed.

tests/test_register_flow.py
quis: I meant you’re missing the word ‘one’ in the failure message.

tests/test_register_flow.py
quis: eg

**test.py**
```python
import pytest

closed = False

def f():
     try:
         pytest.fail("fail")
     finally:
         closed = True
         
assert closed
```

```shell
py.test test.py
>>> E   assert False
```

PR #7, opened 2016-03-15
Use sms inbox

servingUpAces:
- Removed the use of the TwilioRestApi and started using the callback.  
When the Twilio number gets a sms message, it posts the message to https://notify-sms-inbox.herokuapp.com/
The endpoint only keeps a records of the last message posted. The GET request to this endpoint will delete cache of this message as well.
The hope is that the stability of the functional tests will improve. Able to remove a lot of the utils functions.

### DIFF
--- a/sms_inbox.py
+++ b/sms_inbox.py
@@ -13,7 +13,7 @@ def get_nessage():
     if result:
         cache.clear()
         return jsonify({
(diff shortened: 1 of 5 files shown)


mannickutd:
:+1:

PR #13, opened 2016-04-11
Selenisation of tests

ashimali:


tests/pages/pages.py
```diff
@@ -0,0 +1,280 @@
...
+        element.click()
+
+
+class TourPage(BasePage):
+
+    def is_current(self):
+        return self.driver.current_url.endswith('/tour/1')
+
+    def get_me_out_of_here(self):
+        while True:
```
mannickutd: hmm dangerous while loop, is there timeouts for tests?

mannickutd:
Looks good :+1:

tests/pages/pages.py
ashimali: Yeah good point. I'll change this as we know how many links there are.

PR #16, opened 2016-06-20
add delete template functional test

leohemsted:


ashimali:
@leohemsted Just noticed this, and did a bit of rework to match change in test structure. Also link text for editing templates has proliferated, so a hacky tweak for that. 

Anyway if you still want to merge then 👍

PR #18, opened 2016-07-14
test_api_key test

leohemsted:
putting team_api_key test on hold for now before i go stir crazy - but here's the test one. it sets up a test api key, then sends an email and asserts that:
* the notification status is `delivered`
* there is no notification email in the inbox


servingUpAces:
👍

PR #22, opened 2016-09-02
Update staging live tests

servingUpAces:
- Update the staging|live tests to test sending messages via the api. 
- Some refactoring
- Trying to remove calls to check that the message is physically in the inbox
- Removed Procfile and files run on heroku 
- Made the call to get the sms code easier.


tests/staging_live/test_send_notifications_from_csv.py
```diff
@@ -1,36 +1,74 @@
...
+    upload_csv_page.upload_csv(directory, filename)
+    email_body = get_notification_via_api(profile.service_id, profile.email_template_id,
+                                          profile.env, profile.api_key, profile.email)
+    assert "The quick brown fox jumped over the lazy dog" in email_body
+
+
+def send_sms_via_api(client, profile):
+    resp_json = client.send_sms_notification(profile.mobile, profile.sms_template_id)
+    notification_id = resp_json['data']['notification']['id']
+    sleep(5)
```
leohemsted: i don't really like `sleep`s - they're slow in low load, and unreliable in high load. will https://github.com/alphagov/notifications-functional-tests/blob/master/tests/utils.py#L127 work for you ?

leohemsted: yeah after finishing going through, code looks good but i'd try and merge `get_notification_via_api_by_id` and `get_delivered_notification` in utils.py since they do the same thing more or less, except for your smart 'sending' check for dev environments

leohemsted:
👍

PR #30, opened 2016-09-20
Add separate provider delivery tests

imdadahad:
This PR includes additional tests to check if the provider fails for sending notifications via the API and general refactoring.

The Makefile has also been adjusted to include the ability to run these inside a docker container.


servingUpAces:
👍  once the chromedriver is removed.

PR #33, opened 2016-09-22
Add validation to ensure correct notification is returned after csv upload 

imdadahad:
Currently there are no precise checks to ensure that a notification created via a csv is actually correct. As it stands all notifications are returned and then checks occur to verify (as best as possible) that the notification is correct.

This pull requests implements functionality to retrieve the notification id after a csv upload and use this to retrieve the notification via the api and then verify its contents.

### DIFF
--- a/tests/pages/pages.py
+++ b/tests/pages/pages.py
@@ -318,8 +318,6 @@ class UploadCsvPage(BasePage):
         element.click()
 
     def upload_csv(self, directory, path):
-        import pdb
-        pdb.set_trace()
         file_path = os.path.join(directory, 'sample.csv')
         self.file_input_element = file_path
         self.click_send()
--- a/tests/pages/pages.py
+++ b/tests/pages/pages.py
@@ -318,6 +318,8 @@ class UploadCsvPage(BasePage):
         element.click()
 
     def upload_csv(self, directory, path):
+        import pdb
+        pdb.set_trace()
         file_path = os.path.join(directory, 'sample.csv')
         self.file_input_element = file_path
         self.click_send()
--- a/tests/pages/pages.py
+++ b/tests/pages/pages.py
@@ -4,8 +4,9 @@ import shutil
 from selenium.webdriver.common.by import By
 from selenium.webdriver.support.ui import WebDriverWait
 from selenium.webdriver.support import expected_conditions as EC
-from selenium.common.exceptions import NoSuchElementException
+from selenium.common.exceptions import NoSuchElementException, StaleElementReferenceException
 
+from retry import retry
 from config import Config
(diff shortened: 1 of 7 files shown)


tests/pages/pages.py
```diff
@@ -38,6 +39,10 @@
 )
 
 
+class RetryException(Exception):
```
servingUpAces: You can use the RetryException in utils.

tests/test_all_the_things.py
```diff
@@ -16,7 +16,9 @@
     do_verify,
```
servingUpAces: There is some unused imports here.

tests/pages/pages.py
imdadahad: Attempted to import but seems like there's a circular dependency issue as both files are importing each other. Will look into it further.

tests/test_all_the_things.py
imdadahad: Hmm, both used on line 87 and 205.

tests/pages/pages.py
imdadahad: Managed to resolve by moving the RetryException further down (trick found online)

tests/postman.py
```diff
@@ -0,0 +1,49 @@
...
+from tests.utils import create_temp_csv, RetryException
+from notifications_python_client.errors import HTTPError
+
+
+def send_notification_via_api(client, template_id, to, message_type):
+    if message_type == 'sms':
+        # resp_json = client.send_sms_notification(profile.mobile, profile.sms_template_id)
+        resp_json = client.send_sms_notification(to, template_id)
+    elif message_type == 'email':
+        # resp_json = client.send_email_notification(profile.email, profile.email_template_id)
```
servingUpAces: Commented lines can go.

tests/postman.py
```diff
@@ -0,0 +1,49 @@
...
+        resp_json = client.send_email_notification(to, template_id)
+
+    return resp_json['data']['notification']['id']
+
+
+def send_notification_via_csv(profile, upload_csv_page, message_type):
+    if message_type == 'sms':
+        template_id = profile.sms_template_id
+        directory, filename = create_temp_csv(profile.mobile, 'phone number')
+        to = profile.mobile
```
servingUpAces: This variable is not used.

tests/postman.py
imdadahad: Removed

tests/postman.py
imdadahad: Thanks, removed

servingUpAces:
👍  Looks good

PR #36, opened 2016-10-05
Add custom user agent to functional test browsers

quis:
This will make them easy to filter out in Google Analytics.


imdadahad:
👍

PR #39, opened 2016-10-10
Use team key for functional tests

quis:
Services in trial mode can’t create normal keys any more. There’s not even the option on the page (since https://github.com/alphagov/notifications-admin/pull/969). This was causing the tests to fail.


quis:
This is failing because getting a notification is restricted to notifications created with the same key type. Not sure how it was working before. Will need to investigate tomorrow.

quis:
Will pass once https://github.com/alphagov/notifications-api/pull/707 is merged.

PR #40, opened 2016-10-11
Remove api tests from test_all_the_things

leohemsted:
cos they're now based in each client repo, rather than this one


imdadahad:
Included in #41

tests/conftest.py
```diff
@@ -104,7 +104,7 @@ def driver(request):
     elif driver_name == 'chrome':
         options = webdriver.chrome.options.Options()
         options.add_argument("user-agent=Selenium")
-        driver = webdriver.Chrome(chrome_options=opts)
+        driver = webdriver.Chrome(chrome_options=options)
```
imdadahad: yeah I added it in already.

PR #41, opened 2016-10-13
Refactor preview tests

imdadahad:
The following changes have been made:

1. A separate set of tests will run for preview:
    - Register a user, login, invite a user, verify they complete invitation flow
    - Send csv notifications (email & sms) with a research mode service and seeded user

2. The tests that run locally will remain the same, insofar as they will still test everything in 1. but without a research mode service. This is to avoid the pain of performing additional steps to run the functional tests locally 

3. The csv notification tests now include dashboard checks to verify counts are accurate and status is valid

4. General refactoring:
    - Separate retry delay variable
    - Sign out involves actually clicking the sign out link as opposed to a get request to /sign-out
    - Methods moved to utils as they will be shared 

5. For this to run on preview, additional config variables are required in the environment. These changes need to be applied for the tests can execute

### DIFF
--- a/config.py
+++ b/config.py
@@ -16,8 +16,6 @@ class Config(object):
     EMAIL_TRIES = 36
     EMAIL_DELAY = 5
     RETRY_DELAY = 5
-    SMS_TEMPLATE_ID = os.environ[os.environ['ENVIRONMENT'] + '_SMS_TEMPLATE_ID']
-    EMAIL_TEMPLATE_ID = os.environ[os.environ['ENVIRONMENT'] + '_EMAIL_TEMPLATE_ID']
     FUNCTIONAL_TEST_SERVICE_NAME = os.environ['ENVIRONMENT'] + '_Functional Test Service_'
     NOTIFY_SERVICE_ID = os.environ[os.environ['ENVIRONMENT'] + '_NOTIFY_SERVICE_ID']
     NOTIFY_SERVICE_API_KEY = os.environ[os.environ['ENVIRONMENT'] + '_NOTIFY_SERVICE_API_KEY']
--- a/config.py
+++ b/config.py
@@ -15,12 +15,24 @@ class Config(object):
     INVITATION_EMAIL_LABEL = 'invite'
     EMAIL_TRIES = 36
     EMAIL_DELAY = 5
+    RETRY_DELAY = 5
+    SMS_TEMPLATE_ID = os.environ[os.environ['ENVIRONMENT'] + '_SMS_TEMPLATE_ID']
+    EMAIL_TEMPLATE_ID = os.environ[os.environ['ENVIRONMENT'] + '_EMAIL_TEMPLATE_ID']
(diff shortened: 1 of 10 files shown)


tests/conftest.py
```diff
@@ -28,7 +28,38 @@ def __repr__(self):
 @pytest.fixture(scope="session")
 def profile():
     env = os.environ['ENVIRONMENT'].lower()
-    if env == 'staging':
+
+    if env == 'preview':
+        from config import PreviewConfig
+        uuid_for_test_run = str(uuid.uuid1())
```
imdadahad: Updated, thanks

config.py
```diff
@@ -23,6 +26,15 @@ class Config(object):
     VERIFY_CODE_TEMPLATE_ID = '36fb0730-6259-4da1-8a80-c8de22ad4246'
 
 
+class PreviewConfig(Config):
+    SMS_TEMPLATE_ID = os.environ.get('preview_SMS_TEMPLATE_ID')
+    EMAIL_TEMPLATE_ID = os.environ.get('preview_EMAIL_TEMPLATE_ID')
```
leohemsted: this inherits from `Config` so shouldn't need these lines right? [or possibly the ones below either]

config.py
```diff
@@ -15,6 +15,9 @@ class Config(object):
     INVITATION_EMAIL_LABEL = 'invite'
     EMAIL_TRIES = 36
     EMAIL_DELAY = 5
+    RETRY_DELAY = 5
+    SMS_TEMPLATE_ID = os.environ[os.environ['ENVIRONMENT'] + '_SMS_TEMPLATE_ID']
+    EMAIL_TEMPLATE_ID = os.environ[os.environ['ENVIRONMENT'] + '_EMAIL_TEMPLATE_ID']
```
leohemsted: do we need to supply these for running locally? what should they be?

if you run locally without dev_SMS_TEMPLATE_ID defined it keyerrors, which I don't think should happen if you still have non-seeded running by default

tests/conftest.py
```diff
@@ -28,7 +28,38 @@ def __repr__(self):
 @pytest.fixture(scope="session")
 def profile():
     env = os.environ['ENVIRONMENT'].lower()
-    if env == 'staging':
+
+    if env == 'preview':
```
leohemsted: can these if statements be changed to

```
if env == 'preview':
    cfg = PreviewConfig
elif env == 'staging':
    cfg = StagingConfig
elif [...]


return Profile([...])
```

cos this is pretty massive

leohemsted: (thats a can as in "can they feasibly", not "i demand that you change them"

imdadahad: I was going to do this exactly but thought better to leave for next refactor

config.py
imdadahad: RETRY_DELAY is needed but the other two can go 👍

config.py
imdadahad: Based on the above reply, we still need these now and the ones below

imdadahad:
@servingUpAces is also happy after a run-through

PR #45, opened 2016-10-19
Provider test increase retry time

imdadahad:
Previously we were running provider-delivery checks every 5 seconds for a total of 75 seconds.

We want to give more leniency (driven by graphs on dashboard) to check every 20 seconds or so for up-to **4 minutes**.

### DIFF
--- a/config.py
+++ b/config.py
@@ -16,6 +16,8 @@ class Config(object):
     EMAIL_TRIES = 36
     EMAIL_DELAY = 5
     RETRY_DELAY = 5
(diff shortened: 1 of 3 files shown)


leohemsted:
@imdadahad there was a mistake here - the `get_notification_by_id_via_api` call *still* has the 5*15 retry wrapper around it, so this is running  *THAT* retrying code 12 times with a 20 second pause in between each run - so if MMG isn't working, it runs for 12 * ( 20 + ( 5 * 15 ) ) seconds = 20 minutes (plus the execution time of the request 180 times)

PR #46, opened 2016-10-20
Remove emails and make screenshots work across envs

leohemsted:
(previously screenshots would only work if calling run_functional_tests.sh - now work with run_test_script.sh as well)

additionally, remove the imap stuff! We don't use it any more


imdadahad:
👍

PR #48, opened 2016-10-24
Browser logs

leohemsted:
firefox logging may not actually work due to SeleniumHQ/selenium#2698, however upgrading selenium to get that fix (selenium >= 3.0.1) is a prohibitively big change due to differences in how firefox is handled. 

we should probably upgrade to 3.0.1 anyway because upgrading is good, but the changes to how firefox runs (marionette instead of natively) means that's not a straightforward change if we stick with firefox - a bunch of our tests break due to changes in how page changes interact with element finding etc. or we switch to chrome and stuff _might_ change


imdadahad:
👍

scripts/run_functional_tests.sh
```diff
@@ -50,6 +50,8 @@ if [ -d screenshots ]; then
   rm -rfv screenshots
 fi
 mkdir screenshots
+# remove any previous errors
+rm -f browser.log
```
imdadahad: Should also add to .gitignore

leohemsted: `*.log` is already in gitignore

imdadahad: Nice 👍

PR #55, opened 2016-11-07
Record timestamps for tests

imdadahad:
This will enable us to record the start and end time for all tests in each environment.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -3,9 +3,10 @@ import os
 import tempfile
 import csv
 import re
+import logging
 import uuid
-
 import pytest
+
 from datetime import datetime
 from retry import retry
 from notifications_python_client.notifications import NotificationsAPIClient
@@ -23,6 +24,9 @@ from tests.pages import (
     EditEmailTemplatePage,
     VerifyPage
 )
+
+logging.basicConfig(filename='./logs/test_run_{}.log'.format(datetime.utcnow()),
+                    level=logging.INFO)
 
 
 def create_temp_csv(number, field_name):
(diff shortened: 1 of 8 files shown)


tests/utils.py
```diff
@@ -187,3 +189,15 @@ def get_notification_via_api(service_id, template_id, api_key, sent_to):
...
                 sent_to)
     raise RetryException(message)
+
+
+def recordtime(func):
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        print('Start Time: {}'.format(str(datetime.utcnow())))
+        result = func(*args, **kwargs)
+        print('End Time: {}'.format(str(datetime.utcnow())))
```
leohemsted: also is it worth printing out the duration here? what's the need for printing start and end - is it just reconciling log file timestamps or do we want the at-a-glance timing part of it

tests/utils.py
```diff
@@ -187,3 +189,15 @@ def get_notification_via_api(service_id, template_id, api_key, sent_to):
         .format(template_id,
                 sent_to)
     raise RetryException(message)
+
+
+def recordtime(func):
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        print('Start Time: {}'.format(str(datetime.utcnow())))
```
imdadahad: "During test execution any output sent to stdout and stderr is captured." from pytest docs. I think we only want console logs mainly to reconcile the test start / end times with kibana (which took longer than it should have when debugging yesterday)

tests/utils.py
imdadahad: Yeah good idea, ideally we want as much info as we can get in both scenarios (start time/end time). Duration may be useful too although not key, as objective was to reconcile logs

PR #70, opened 2017-02-01
Fix issue causing exception on verification

imdadahad:
This fixes an issue where our functional tests fails during 2fa, despite having verified successfully.

During the 2fa stage, a `TimeoutException` is raised occasionally due to verification _apparently_ not being completed. However after checking we can see verification succeeded. Until we can dig deep to find the real issue, this adds a explicit check to see if we verified and proceed on with the test if so.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -10,7 +10,7 @@ import pytest
 from datetime import datetime
 from retry import retry
 from notifications_python_client.notifications import NotificationsAPIClient
-from selenium.common.exceptions import NoSuchElementException
+from selenium.common.exceptions import NoSuchElementException, TimeoutException
 from selenium.webdriver.support.ui import WebDriverWait
 
 from config import Config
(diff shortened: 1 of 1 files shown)


tests/test_utils.py
```diff
@@ -76,9 +76,11 @@ def __call__(self, driver):
...
                 verify_page.verify(verify_code)
                 driver.find_element_by_class_name('error-message')
-            except NoSuchElementException:
-                #  We were able to verify as there was no error message
-                return True
+            except (NoSuchElementException, TimeoutException) as e:
+                #  In some cases a TimeoutException is raised even if we have managed to verify.
+                #  For now, check explicitly if we 'have verified' and if so move on.
+                if 'two-factor' not in driver.current_url:
+                    return True
```
leohemsted: as discussed, don't think we need the `if 'two-factor' not in driver.current_url` check here:

* TimeoutException is only caused by trying to find verify code, but the box isn't there - the only time that is the case is when we're on the next page (logged in, selenium doesn't know that yet)
  - It could be caused by general slowness (which is a problem we'd want func tests to flag), but if that's the case we'd probably catch it earlier on anyway
* NoSuchElementException can be caused by two things: 
    1. being logged in - we should return True
    2. not received the first error code yet - we should retry

in case of NoSuchElement i don't think we can identify which of those two cases it is, I'd suggest we return True every time, and then if we run into problems where we should have retried but thought we were logged in, they'll be flagged up very quickly.

PR #71, opened 2017-02-09
Add Jenkins build id to message body for functional tests

imdadahad:
This will add the Jenkins build id (pulled from the environment) to the the message body in the functional tests.

## How it works

1. **Dev** - `test_all_the_things.py` creates and uses templates to run the tests. They are triggered on submission of a pull request on the functional tests. The code for creating templates has been modified to include the `build_id`.

2. **Preview, Staging and Production** - Templates that will include the Jenkins build id have been created on preview, staging and production. These environment variables have been set and deployed in Jenkins [(229ead1)](https://github.com/alphagov/notifications-aws/commit/229ead1ed69940db1091c52564263a64982ed707). These tests will use the new templates that include the build id. 

The code has been refactored so that creation of a CSV will include an extra column for the build id and also the API calls will include the build id in the personalisation.

### Why the need for new template IDs?

This was to avoid our tests breaking. If the existing template ids were modified in Jenkins before any code changes were merged, our tests would break as they wouldn't yet include the personalisation required by the API. As a result it was essential to create new templates to avoid this. 

## How to test

Test on development as usual. You won't need to mess around with any environment variables.

To test against preview, staging or production set these two variables for each environment in your `.sh`. E.g:
`live_JENKINS_BUILD_SMS_TEMPLATE_ID`
`live_JENKINS_BUILD_EMAIL_TEMPLATE_ID`

These will be added to the shared stored for convenience.


tests/admin/test_admin_with_seeded_user.py
```diff
@@ -34,7 +34,12 @@ def test_registration_and_invite_flow(driver, profile, base_url):
 def test_send_csv(driver, profile, login_seeded_user, seeded_client, message_type):
     dashboard_page = DashboardPage(driver)
     dashboard_page.go_to_dashboard_for_service()
-    template_id = profile.email_template_id if message_type == 'email' else profile.sms_template_id
+
+    template_id = None
+    if message_type == 'email':
+        template_id = profile.jenkins_build_email_template_id
+    elif message_type == 'sms':
+        template_id = profile.jenkins_build_sms_template_id
```
quis: Here’s a more concise pattern that we use elsewhere for this kind of thing:
```python
template_id = {
    'email': profile.jenkins_build_email_template_id,
    'sms': profile.jenkins_build_sms_template_id,
}.get(message_type)
```

tests/pages/rollups.py
```diff
@@ -68,8 +68,8 @@ def get_service_templates_and_api_key_for_tests(driver, test_profile):
...
 
     test_profile.service_id = service_id
-    test_profile.email_template_id = email_template_id
-    test_profile.sms_template_id = sms_template_id
+    test_profile.jenkins_build_email_template_id = email_template_id
+    test_profile.jenkins_build_sms_template_id = sms_template_id
     test_profile.api_key = api_key
 
-    return {'service_id': service_id, 'email_template_id': email_template_id, 'sms_template_id': sms_template_id, 'api_key': api_key}  # noqa
+    return {'service_id': service_id, 'api_key': api_key}  # noqa
```
quis: Can lose the `#noqa` on this line now.

tests/admin/test_admin_with_seeded_user.py
imdadahad: Nice 👍

PR #74, opened 2017-02-27
Add build id into docker env

imdadahad:
This fixes the issue where the Jenkins build id is not shown in our emails/sms. This is because the build id was missing from the docker env and needs to be added explicitly.

### DIFF
--- a/scripts/generate_docker_env.sh
+++ b/scripts/generate_docker_env.sh
@@ -40,3 +40,5 @@ env_vars=(
 for env_var in "${env_vars[@]}"; do
     echo "${ENVIRONMENT}_${env_var}=${!env_var}" >> docker.env
 done
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -105,6 +106,7 @@ define run_test_container
 		-e https_proxy="${HTTPS_PROXY}" \
 		-e HTTPS_PROXY="${HTTPS_PROXY}" \
 		-e NO_PROXY="${NO_PROXY}" \
+		-e BUILD_ID="${BUILD_ID}" \
 		--env-file docker.env \
```
leohemsted: any reason why you don't want to put it in the docker.env instead with the other vars?

imdadahad: Added and squashed out 👍

PR #79, opened 2017-06-23
Small app to test the inbound sms flow.

servingUpAces:
Contains an endpoint to receive the pushed inbound sms from the notify api.


receive_inbound_sms/scripts/run_app.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+python3 app.py runserver
```
leohemsted: Few things:

* Think this should be `application.py` rather than `app.py`
* runserver appears to be a command that we add ourselves, see https://github.com/alphagov/notifications-api/blob/master/application.py#L12 - does this work without?
* we aren't using $PORT, that is passed in to the script. We need to expose it on the port that PaaS knows about or we won't be able to access it

PR #90, opened 2017-10-17
Refactor code to work with admin

kentsanggds:
Existing tests when run under `dev` environment were broken as they were not updating in line with admin.

This update should ensure that tests can be run in a local `dev` environment and on PRs, there might be a wider discussion on what functional tests should be run but this PR is just focused on getting them back in line and working with the current admin pages.


leohemsted:
looks like `do_edit_and_delete_email_template(driver)` is the only bit of this that is different to the preview functional tests - lets kill test_all_the_things and copy that bit to preview

tests/test_utils.py
```diff
@@ -204,12 +211,12 @@ def get_notification_via_api(service_id, template_id, api_key, sent_to, attempt_
         to = notification['to']
         status = notification['status']
         if t_id == template_id and to == sent_to and status in ['sending', 'delivered']:
+            print('matching notification', notification)
```
leohemsted: think we should kill this print now. Lets keep the 2fa one above though.

kentsanggds: ok

PR #103, opened 2017-11-15
Replace pep8 references with flake8 references

klssmith:
- We now use flake8

### DIFF
--- a/scripts/run_functional_tests.sh
+++ b/scripts/run_functional_tests.sh
@@ -27,7 +27,7 @@ if [ -d venv ]; then
   source ./venv/bin/activate
 fi
 
-pep8 --exclude=venv .
+flake8 .
 display_result $? 1 "Code style check"
 
 # default env to master (i.e. preview)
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -1,2 +1,2 @@
-[pep8]
+[flake8]
```
quis: This is configured in [.flake8](https://github.com/alphagov/notifications-functional-tests/blob/4b22e3b4b5ebf9f8263d417af5814492796d258d/.flake8) now instead – this file should be deleted.

klssmith: Thanks - added a commit to delete it

scripts/run_functional_tests.sh
```diff
@@ -27,7 +27,7 @@ if [ -d venv ]; then
   source ./venv/bin/activate
 fi
 
-pep8 --exclude=venv .
+flake8 --exclude=venv .
```
quis: The `--exclude` shouldn’t be needed any more either cos that’s also configured in .flake8

PR #133, opened 2018-04-12
Update Makefile

leohemsted:
lower case stuff too


leohemsted:
not needed it turns out

PR #134, opened 2018-04-17
Add tests for successfully sent precompiled letters

richardc0:
Updated to allow successfully sent precompiled letters to be tested so that when we go live with precompiled we have some assurance that the functionality it working.

- Test that we can send a precompiled letter via API
- Added checks that letter appears in API integration / message log


tests/postman.py
```diff
@@ -19,6 +19,13 @@ def send_notification_via_api(client, template_id, to, message_type):
     return resp_json['id']
 
 
+def send_precompiled_letter_via_api(client, pdf_file):
+    reference = 'Build: {}'.format(os.getenv('BUILD_ID', 'No build id'))
```
leohemsted: should probably involve uuid_for_test_run so that if we rebuild we won't pick up the ref from another build

PR #135, opened 2018-04-23
Add precompiled letters tests

richardc0:
Add test for precompiled letters to test some basic functionality using the message log as validation, tests should check for a delivered message and a virus_scan_failed message:

* Added API tests which test the API and its return only for good and virus_scan_failed PDFs
* Added user tests which check the the API message log for occurrences of the notification and also check that the status of the notification is as expected
* Added to locators for the message log to get the various items from the descriptive list
* Added util functions to get data from the various lists etc


requirements.txt
```diff
@@ -8,4 +8,4 @@ pytest==3.5.0
 retry==0.9.2
 selenium==3.11.0
 
-git+https://github.com/alphagov/notifications-python-client.git@4.5.0#egg=notifications-python-client==4.5.0
+git+https://github.com/alphagov/notifications-python-client.git@4.8.1#egg=notifications-python-client==4.8.1
```
servingUpAces: Latest version is now 4.8.2

PR #141, opened 2018-05-10
Separate env files

leohemsted:
## Why I've done:

Frustration points:
* Long variable names (eg `profile.notify_research_service_api_test_key`)
* Mix of `Config.SOME_VARIABLE` and `profile.some_variable` used throughout.
* often passing `profile` fixture around to lots of helper functions
* confusing debugging due to `os.environ.get` leading to some variables being `None`
* lengthy config file full of duplication - every line has a `os.environ.get(os.environ.get('ENVIRONMENT') + '_SOME_VARIABLE')`
* lots of duplication of env var busy-work between config.py and conftest.py
* hard to tell which variables are associated with which service (think notify vs seeded etc)

## What I've done:

* no longer provide api/admin URLs (they're not secret)
* remove a couple of other things we don't use
* config is now a dict rather than a class, populated through an autouse fixture. This means there are no more `environ.get`. it will load the vars needed for that test run, if they aren't there it'll keyerror.
* remove the profile fixture - now you should say `from config import config`.
* remove prefixes from environment variables. Put your vars in separate
  files. For example, have an `environment_preview.sh`, an `environment_dev.sh`, etc etc. *YOU WILL NEED TO UPDATE YOUR LOCAL ENVIRONMENT FILES*
* can't use curl to print out the build now because we don't know the base url until python time - so a little in-line python script is used
* simplify generate_docker_env - now there's some duplication but it's a much simpler bash script that just pastes stuff into the file rather than looping through vars.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,7 @@ test-providers: venv ## Run tests
 	su -c '/var/project/scripts/run_test_script.sh /var/project/tests/provider_delivery/' hostuser
 
 .PHONY: test-document-download
-test-document-download: venv ## Run documnet-download-api tests
+test-document-download: venv ## Run document-download-api tests
 	su -c '/var/project/scripts/run_test_script.sh /var/project/tests/document_download/' hostuser
 
 .PHONY: generate-env-file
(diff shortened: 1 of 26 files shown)


Makefile
```diff
@@ -94,13 +94,12 @@ build-with-docker: prepare-docker-runner-image ## Build inside a Docker containe
 		make build
 
 define run_test_container
+	[ ${ENVIRONMENT} != "dev" ] # Cannot run in docker against development
```
servingUpAces: does that comment still apply?

leohemsted: yeah - it can't talk to your local environment (because `localhost` is the local of the docker). Sakis and I tried to get it to work briefly but decided it wasn't worth the effort - you don't see what's going on because it's running inside docker, which is the main reason to run functional tests locally imo.

servingUpAces: fair enough

PR #183, opened 2018-10-08
[do not merge] Remove include_jobs parameter from test

klssmith:
The notifications-api endpoint now returns notifications sent by a job (see https://github.com/alphagov/notifications-api/pull/2139), so there is no longer the option to pass the `include_jobs` parameter in.

[Pivotal story](https://www.pivotaltracker.com/story/show/160757291)


klssmith:
We've decided not to include jobs in the api response

PR #198, opened 2018-11-29
Added a test for basic view and view for a user with all permissions.

servingUpAces:
There is now a test that checks the basic view user goes to the templates page and that the navigation list is correct. 
There is another test for a user with all permissions, expecting them to be on the dashboard page and has the correct navigation list.

### DIFF
--- a/scripts/run_functional_tests.sh
+++ b/scripts/run_functional_tests.sh
@@ -64,7 +64,7 @@ case $ENVIRONMENT in
       ;;
     *)
       # dev, preview
-      py.test -x -v tests/functional/preview_and_dev/test_registration_and_invite.py
+      py.test -x -v tests/functional/preview_and_dev/
       ;;
 esac
 
--- a/scripts/run_functional_tests.sh
+++ b/scripts/run_functional_tests.sh
@@ -64,7 +64,7 @@ case $ENVIRONMENT in
       ;;
     *)
       # dev, preview
-      py.test -x -v tests/functional/preview_and_dev/
+      py.test -x -v tests/functional/preview_and_dev/test_registration_and_invite.py
       ;;
 esac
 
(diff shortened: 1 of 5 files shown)


scripts/run_functional_tests.sh
```diff
@@ -64,7 +64,7 @@ case $ENVIRONMENT in
       ;;
     *)
       # dev, preview
-      py.test -x -v tests/functional/preview_and_dev/
+      py.test -x -v tests/functional/preview_and_dev/test_registration_and_invite.py
```
klssmith: Was this change to just run one test supposed to be committed?

servingUpAces: oops - I thought I took that out.

tests/test_utils.py
```diff
@@ -170,9 +173,12 @@ def do_user_can_invite_someone_to_notify(driver):
     dashboard_page.go_to_dashboard_for_service(service_id)
 
     assert dashboard_page.h2_is_service_name(config['service_name'])
-
-    dashboard_page.sign_out()
-    dashboard_page.wait_until_url_is(config['notify_admin_url'])
+    if basic_view:
+        dashboard_page.is_for_basic_view()
+        dashboard_page.sign_out()
+        dashboard_page.wait_until_url_is(config['notify_admin_url'])
```
klssmith: Do we need the lines to sign out and wait until the URL changes? It's not being done in the `else` (when we're not in basic view) and it looks like this is the last test that runs anyway

tests/pages/pages.py
```diff
@@ -339,6 +344,16 @@ def get_template_message_count(self, template_id):
 
         return int(element.text)
 
+    def is_for_basic_view(self):
```
klssmith: I would consider moving `is_for_basic_view` and `is_for_view_with_all_permissions` to the file where they're used. It makes it more obvious that these functions are tests with assertions (because we don't have any other assertions in this file - everything else here is to do with finding elements or clicking things)

tests/test_utils.py
servingUpAces: Unfortunately, I need to do this because the user with permissions invites a user with basic view, so the first user needs to stay logged in.

PR #227, opened 2019-02-21
Pending is a sending status

leohemsted:
(this'll mean the functional tests don't fail if the notification is in pending for a long time)

### DIFF
--- a/tests/functional/staging_and_prod/test_admin.py
+++ b/tests/functional/staging_and_prod/test_admin.py
@@ -6,10 +6,9 @@ from tests.pages import UploadCsvPage
 from tests.postman import (
     send_notification_via_csv,
     get_notification_by_id_via_api,
-    NotificationStatuses
 )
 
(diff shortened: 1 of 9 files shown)


tests/functional/staging_and_prod/test_admin.py
```diff
@@ -6,14 +6,14 @@
...
-    NotificationStatuses
 )
 
-from tests.test_utils import assert_notification_body, recordtime
+from tests.test_utils import assert_notification_body, recordtime, NotificationStatuses
 
 
 @recordtime
 def test_admin(driver, client, login_user):
+    raise NotImplementedError
```
CrystalPea: do I understand correctly that we are disabling this test because it fails if our providers are slow but it is not what it was meant to test? If yes, do we have another test for sending notifications through jobs?

leohemsted: whoops, that was a local change that i didn't mean to commit. squashed out

PR #233, opened 2019-03-28
Folder permissions

CrystalPea:
Part of this story: https://www.pivotaltracker.com/story/show/164632434

### DIFF
--- a/tests/functional/preview_and_dev/test_seeded_user.py
+++ b/tests/functional/preview_and_dev/test_seeded_user.py
@@ -373,6 +373,11 @@ def test_template_folder_permissions(driver, login_seeded_user):
     edit_team_member_page.uncheck_folder_permission_checkbox(folder_names[1])
     edit_team_member_page.click_save()
 
+    # check if permissions saved correctly
+    dashboard_page.click_team_members_link()
+    team_members_page.click_edit_team_member()
+    assert not edit_team_member_page.is_checkbox_checked(folder_names[1])
+
     # log out
     dashboard_page.sign_out()
     # log in as that colleague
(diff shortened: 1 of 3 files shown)


tests/functional/preview_and_dev/test_seeded_user.py
```diff
@@ -334,6 +338,84 @@ def test_creating_moving_and_deleting_template_folders(driver, login_seeded_user
     assert template_name not in [x.text for x in driver.find_elements_by_class_name('message-name')]
 
 
+def test_template_folder_permissions(driver, login_seeded_user):
+    folder_names = [
+        'test-parent-folder {}'.format(uuid.uuid4()),
+        'test-child-folder {}'.format(uuid.uuid4()),
+        'test-grandchild-folder {}'.format(uuid.uuid4()),
```
leohemsted: do we want these to have the same uuid? (as a kind of "test run" sort of thing)

not sure it makes a huge difference, just wonder if it'll make it easier to debug/search if a test fails

tests/pages/pages.py
```diff
@@ -586,6 +628,13 @@ def click_invite_user(self):
         element = self.wait_for_element(TeamMembersPage.invite_team_member_button)
         element.click()
 
+    def click_edit_team_member(self, email=None):
```
leohemsted: we never call this without passing `email` in, so i think we can get rid of this custom url

CrystalPea: ahh, you are right, probably it was a leftover 😅. Fixed now :)

tests/functional/preview_and_dev/test_seeded_user.py
CrystalPea: Good suggestion, I changed them to have common id now :)

PR #247, opened 2019-05-28
Update tests to handle email reply-to address verification feature

CrystalPea:
Needs code from those PRs to pass:
API: https://github.com/alphagov/notifications-api/pull/2511
Admin: https://github.com/alphagov/notifications-admin/pull/2974
(and those PRs need this update here to pass functional tests :D )


tests/pages/pages.py
```diff
@@ -224,6 +224,8 @@ def is_current(self):
 
     def add_service(self, name):
         self.service_input = name
+        radio = self.wait_for_invisible_element(AddServicePageLocators.CENTRAL_GOVT_RADIO)
```
quis: This will fail on preview because the preview functional test service belongs to an organisation. We should revert this PR, then it will work in both situations: https://github.com/alphagov/notifications-functional-tests/pull/240

CrystalPea: Cool, thanks for that, I was pretty confused about it.

Here's the revert PR:
https://github.com/alphagov/notifications-functional-tests/pull/248

tests/test_utils.py
```diff
@@ -271,6 +271,8 @@ def wrapper(*args, **kwargs):
 
 
 def do_user_can_add_reply_to_email_to_service(driver):
+    if not config["env"] == "preview":
```
leohemsted: good habit to use `!=` rather than `not .. == ..`

tests/test_utils.py
```diff
@@ -31,9 +31,9 @@
 
 jenkins_build_id = os.getenv('BUILD_ID', 'No build id')
 
-email_address = "notify@digitial.cabinet-office.gov.uk"
-email_address2 = "notify+1@digitial.cabinet-office.gov.uk"
-email_address3 = "notify+2@digitial.cabinet-office.gov.uk"
+email_address = os.environ['NOTIFY_RESEARCH_EMAIL_REPLY_TO']
+email_address2 = os.environ['NOTIFY_RESEARCH_EMAIL_REPLY_TO_2']
+email_address3 = os.environ['NOTIFY_RESEARCH_EMAIL_REPLY_TO_3']
```
leohemsted: any reason why you're reading from os.environ here rather than from config.py? also, what happens if you don't have these environment variables set (locally)?

CrystalPea: thanks, I changed it now :)

PR #261, opened 2019-07-12
add SELENIUM_DRIVER to the dockerfile

leohemsted:
so that we can configure what browser we use when running in docker


leohemsted:
wait it was already in the makefile

PR #269, opened 2019-08-07
Run doc download tests when testing antivirus

andy-paine:
- The existing antivirus test only tests celery but we want to test the
  antivirus API too
- The doc download test should hit API

### DIFF
--- a/scripts/run_antivirus_functional_test.sh
+++ b/scripts/run_antivirus_functional_test.sh
@@ -47,6 +47,6 @@ mkdir logs
 echo "Running $ENVIRONMENT tests"
 display_status $ENVIRONMENT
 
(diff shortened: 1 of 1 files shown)


scripts/run_antivirus_functional_test.sh
```diff
@@ -47,6 +47,6 @@ mkdir logs
 echo "Running $ENVIRONMENT tests"
 display_status $ENVIRONMENT
 
-py.test -k test_send_precompiled_letter_with_virus_notification_via_api -x -v
+py.test -k 'test_send_precompiled_letter_with_virus_notification_via_api or test_document_download' -x -v
```
leohemsted: what i had no idea you could do this this is wild

PR #276, opened 2019-09-24
BAU: Update chrome public key

andy-paine:
- Google have updated their package signing key so we need to re-import
  it here


andy-paine:
test this please

PR #277, opened 2019-09-26
improve service renaming test

leohemsted:


tests/pages/pages.py
```diff
@@ -870,14 +870,12 @@ def send_to_myself(self, message_type):
 
 
 class ServiceSettingsPage(BasePage):
-    def check_service_name(self, expected_name=None):
+    def check_service_name(self, expected_name):
```
idavidmcdonald: Could you give a little context of what the problem was with this method before and therefore why you decided to change it's behaviour please.

PR #279, opened 2019-10-02
New Dockerfile that will be used by the concourse build.

servingUpAces:
The new Dockerfile is much more condensed and doesn't need alot of the extra stuff the other Dockerfile and build uses.
Once we are migrated to concourse we can remove the existing Docker folder plus Makefile.

We also a headless option for the chrome driver. The tests don't pass on concourse without it and seem to still pass on Jenkins. Documentation is a little spare around the headless mode (https://www.chromestatus.com/features/5678767817097216), however I can't see any immediate downsides of running our tests in it.

Concourse success - https://cd.gds-reliability.engineering/teams/notify/pipelines/notify/jobs/functional-tests-preview/builds/132

## Without setting the window size
image

## With setting the window size
image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,5 +5,6 @@ ADD requirements.txt requirements.txt
 RUN pip install -r requirements.txt
 ADD tests/ tests/
 ADD config.py config.py
+ADD scripts/ scripts/
 RUN mkdir logs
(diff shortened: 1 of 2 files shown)


idavidmcdonald:
Don't merge this yet please anyone

PR #292, opened 2019-11-27
Update Python to 3.7 when running in Docker

quis:
We can’t upgrade to the newest version of Pytest while we’re running on Python 3.4. We’ll be running on Python 3.7 once we move to Concourse, which uses the other Dockerfile.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.4-slim-jessie
+FROM joyzoursky/python-chromedriver:3.7-selenium
 
 ARG HTTP_PROXY
 ARG HTTPS_PROXY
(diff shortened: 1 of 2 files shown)


idavidmcdonald:
We have a lot of the functional tests still running on Jenkins that haven't been migrated (mostly the smoke tests in production). Will take a few weeks to migrate that stuff so this won't be able to be merged until then.

idavidmcdonald:
Also your commit message mentions upgrading to 3.6 rather than 3.7

quis:
Ah, not worth it then. Let’s wait for concourse.

idavidmcdonald:
I think I'd like to reopen this.

Pros:
- Both CIs will be running the same version of python meaning we can upgrade to pytest 5
- Currently PRs are not being checked that they can run on Jenkins, just concourse. This means that a functional tests PR might pass it's test and we merge it, only to find out it breaks on Jenkins (this would be the case for the open PR upgrading to pytest 5). Merging this PR will mean we don't merge PRs that might break things without a developer knowing

Cons:
Not sure if this PR itself passes on Jenkins. Will reopen to investigate if that is the case and undisable jenkins for PRs to see

PR #309, opened 2020-02-03
Update selectors for GOVUK Frontend markup

tombye:
GOVUK Frontend was added in this pull request:

https://github.com/alphagov/document-download-frontend/pull/53

This updates the selectors for Document Download
Frontend pages so they can work with the new
markup this adds.

These changes now work before and after the changes to Document Download
Frontend.

### DIFF
--- a/tests/pages/pages.py
+++ b/tests/pages/pages.py
@@ -1035,12 +1035,16 @@ class DocumentDownloadLandingPage(BasePage):
     continue_button = (By.CSS_SELECTOR, 'a.govuk-button')
 
     def get_service_name(self):
-        element = self.wait_for_element((By.CSS_SELECTOR, '.govuk-body > p:first-child'))
+        element = self.wait_for_element((By.CSS_SELECTOR, 'main p:first-of-type'))
 
         return element.text.partition(' sent you ')[0]
 
     def go_to_download_page(self):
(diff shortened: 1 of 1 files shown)


quis:
You need to `try`/`catch` and fallback to the old selector so that this will work with both the new code (once it’s deployed) and the old code (for the time between merging this and merging the code change).

Similar example here:
https://github.com/alphagov/notifications-functional-tests/pull/238/files#diff-eadc29726aea68cd345824b7a9045fe8R222-R226

tests/pages/pages.py
```diff
@@ -1032,15 +1032,19 @@ def get_message(self, content):
...
-        element = self.wait_for_element((By.CSS_SELECTOR, 'h1+p'))
+        element = self.wait_for_element((By.CSS_SELECTOR, 'main p:first-of-type'))
 
         return element.text.partition(' sent you ')[0]
 
     def go_to_download_page(self):
-        button = self.wait_for_element(self.continue_button)
+        try:
+            button = self.wait_for_element(self.continue_button)
+        except:
```
quis: Shouldn’t blanket catch all exceptions, but if it works then fine for now.
```suggestion
        except NoSuchElementException:
```

PR #312, opened 2020-02-13
Update locators for the new button style

klssmith:
We are now switching notifications-admin to use the govuk-frontend style of buttons, which means that most of the button locators need to be updated. We can't find the buttons using the new class name of `govuk-button` because the cookie banner buttons also have this class and they are present in the HTML even when they aren't visible on the page. Finding the button using the CSS selector of `main button.govuk-button` works in most cases since the cookie banner buttons are not in the `main` element.

This PR will fail until https://github.com/alphagov/notifications-admin/pull/3290 is merged, but can be tested locally with that branch.

### DIFF
--- a/tests/pages/pages.py
+++ b/tests/pages/pages.py
@@ -459,10 +459,10 @@ class ShowTemplatesPage(BasePage):
     def _select_template_type(self, type):
         # wait for continue button to be displayed - sticky nav has rendered properly
         # we've seen issues
-        continue_element = self.wait_for_element(self.continue_button)
         radio_element = self.wait_for_invisible_element(type)
-
         self.select_checkbox_or_radio(radio_element)
+
+        continue_element = self.wait_for_element(self.continue_button)
         continue_element.click()
 
     def select_email(self):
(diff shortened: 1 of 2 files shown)


tests/pages/pages.py
```diff
@@ -459,10 +459,10 @@ def click_template_by_link_text(self, link_text):
     def _select_template_type(self, type):
         # wait for continue button to be displayed - sticky nav has rendered properly
         # we've seen issues
-        continue_element = self.wait_for_element(self.continue_button)
         radio_element = self.wait_for_invisible_element(type)
-
         self.select_checkbox_or_radio(radio_element)
+
+        continue_element = self.wait_for_element(self.continue_button)
```
idavidmcdonald: Any reasoning for this reordering of things here? I didn't see anything in your commit message that quite explained it

klssmith: I've split that change out into https://github.com/alphagov/notifications-functional-tests/pull/312/commits/88d7fe7e071f1eff6b3fdd35581741ea96356fb2 and put an explanation in the commit message. We were seeing random failures at that point in the code where the continue button couldn't be found and it seemed best to try and locate an element at the point we want to use it, since locating the element moves the screen to that element.

PR #314, opened 2020-02-20
Add test for password reset flow, also get rid of redundant code

CrystalPea:
The new test test password reset flow. A positive side-effect is that it updates `email_access_validated_at` field for our seeded user.

Connected to that: should I also copy this test to staging_and_prod folder, so seeded user `email_access_validated_at` field is also updated in staging and production?

### DIFF
--- a/tests/functional/preview_and_dev/test_email_auth.py
+++ b/tests/functional/preview_and_dev/test_email_auth.py
@@ -17,7 +17,7 @@ def test_email_auth(driver):
 
 
 @recordtime
-def test_reset_password(driver):
+def test_reset_forgotten_password(driver):
     email, password = get_email_and_password(account_type='seeded')
     sign_in_page = SignInPage(driver)
     sign_in_page.get()
@@ -25,7 +25,7 @@ def test_reset_password(driver):
     sign_in_page.click_forgot_password_link()
 
     forgot_password_page = ForgotPasswordPage(driver)
-    assert forgot_password_page.is_page_title("Forgot your password?")
+    assert forgot_password_page.is_page_title("Forgotten your password?")
     forgot_password_page.input_email_address(email)
     forgot_password_page.click_continue()
     assert forgot_password_page.is_page_title("Check your email")
--- a/tests/functional/preview_and_dev/test_email_auth.py
+++ b/tests/functional/preview_and_dev/test_email_auth.py
@@ -1,7 +1,9 @@
 from config import config
-from tests.test_utils import recordtime
+from tests.test_utils import do_email_verification, recordtime
 
 from tests.pages.rollups import sign_in_email_auth
+from tests.pages import BasePage, ForgotPasswordPage, NewPasswordPage, SignInPage
+from tests.pages.rollups import get_email_and_password
 
 
 @recordtime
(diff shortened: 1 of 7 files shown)


tests/functional/preview_and_dev/test_email_auth.py
```diff
@@ -10,3 +12,32 @@ def test_email_auth(driver):
     sign_in_email_auth(driver)
     # assert url is research mode service's dashboard
     assert driver.current_url == config['notify_admin_url'] + '/services/{}'.format(config['service']['id'])
+    base_page = BasePage(driver)
+    base_page.sign_out()
+
+
+@recordtime
+def test_reset_password(driver):
```
idavidmcdonald: Optional: You could rename this `test_reset_forgotten_password` just so it's very clear this is the flow to reset a forgotten password as opposed to just changing your password after you've logged in

tests/pages/locators.py
```diff
@@ -4,6 +4,7 @@
 class CommonPageLocators(object):
     NAME_INPUT = (By.NAME, 'name')
     EMAIL_INPUT = (By.NAME, 'email_address')
+    NEW_PASSWORD_INPUT = (By.NAME, 'new_password')
```
idavidmcdonald: I'm not sure if I'd put this one in the common page locators section given it feels like it would be pretty rare, I'm assuming only used at most on the change your password form and the reset your forgotten password form

tests/test_utils.py
```diff
@@ -98,14 +98,22 @@ def do_verify(driver):
...
 def do_email_auth_verify(driver):
+    do_email_verification(
+        driver,
+        config['notify_templates']['email_auth_template_id'],
+        config['service']['email_auth_account']
+    )
+
+
+@retry(RetryException, tries=config['verify_code_retry_times'], delay=config['verify_code_retry_interval'])
+def do_email_verification(driver, template_id, email_address):
```
idavidmcdonald: Good refactor here to keep this DRY

tests/pages/pages.py
```diff
@@ -235,11 +236,7 @@ def register(self):
         self.email_input = config['user']['email']
         self.mobile_input = config['user']['mobile']
         self.password_input = config['user']['password']
-        self.click_continue_button()
-
-    def click_continue_button(self):
-        element = self.wait_for_element(RegistrationPage.continue_button)
-        element.click()
+        self.click_continue()
```
idavidmcdonald: Oh the other thing I forgot to say in my previous review was it would have been nice to see this refactoring of existing functionality done in a separate commit as it makes reviewing easier and makes it clearer what the changes are and why. Don't worry, definitely not the end of the world but always good when you can to do refactoring and adding/editing functionality in separate commits.

tests/functional/preview_and_dev/test_email_auth.py
CrystalPea: Nice one, will do

tests/pages/pages.py
CrystalPea: That's a good advice, I will try to separate my tasks better in the future :)

tests/pages/locators.py
CrystalPea: yeah, you're probably right. And also about password reset not being something that we want happening ever 10 minutes 😅 . We will have to  think of a different strategy for that then.

leohemsted:
we could probs change staging/live functional test user to log in via email rather than sms to avoid the 90 days problem

CrystalPea:
@leohemsted that's a cool idea, I'm all for it :)

PR #327, opened 2020-04-17
Letter smoke tests

idavidmcdonald:
Please review commit by commit.

It required me logging into staging and production functional test user accounts and 
- turning on letters for both services
- setting up test API keys which I have added to the creds repo in https://github.com/alphagov/notifications-credentials/pull/168

Have run successfully from my local against staging and production

image
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,74 +1,54 @@
 # notifications-functional-tests
-Functional tests for Notification applications
 
-# Running the tests
+The tests are:
+ 
+- Selenium web driver tests of the Notify user interface
+- tests of the API using the [python API client](https://github.com/alphagov/notifications-python-client).
 
-## On a local dev machine
+## Running the tests from your local dev machine
 
-The majority of tests that are used in local development and also run on master build on Jenkins (running against preview environment) are Selenium web driver tests.
+```shell
+brew cask install chromedriver # needs to be >= v2.32
+```
 
-There is an order dependency in the main tests. The registration test must run before any of the other tests as a new user account created for each test run. That user account is used for all later browser based tests. Each test run will first register a user account using the configured FUNCTIONAL_TEST_EMAIL. The email account will have random characters added so that we do not have uniqueness issues with the email address of registered user.
+To run locally you need to populate a `.gitignore` and `environment.sh` file with the relevant values.
 
-In the main suite there are also tests that that directly use the [python client](https://github.com/alphagov/notifications-python-client) for the notifications api. The client tests require an existing user, service, api key and both email and sms templates.
-
-To run locally you need to populate a `.gitignore` and `environment.sh` file with the relevant values. On Jenkins the environment variables are set in the build settings page.
-
-Note: Your local celery must be run with `ANTIVIRUS_ENABLED=1` set in the environment for the test_view_precompiled_letter_message_log_virus_scan_failed test to work
-
-## Local environment file
-
-- Make sure `Notifications Admin`, `Notifications Template Preview`, `Notifications API`, `Notifications API celery` and `Document Download API` are running locally.
-- to run locally, `source environment_local.sh`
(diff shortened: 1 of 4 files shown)


tests/functional/staging_and_prod/notify_api/test_notify_api_letter.py
```diff
@@ -0,0 +1,33 @@
...
+
+    notification_id = send_precompiled_letter_via_api(
+        reference,
+        seeded_client_using_test_key,
+        BytesIO(base64.b64decode(correct_letter))
+    )
+
+    notification = retry_call(
+        get_notification_by_id_via_api,
+        fargs=[seeded_client_using_test_key, notification_id, NotificationStatuses.RECEIVED],
```
leohemsted, line 28: what does this actually test? Which steps are guaranteed to have run by the time this test has finished?

README.md
```diff
@@ -41,19 +40,15 @@ Running the tests
...
-All the relevant environment variables are setup in the build settings on Jenkins for this repo.
-
-
-### Staging and Live builds
+### Staging and Production
 
-To run against staging and live environments a seeded user account on each of those environments has been created. In addition a service for the user has been created as well as an email and sms template created.
+To run against staging and production environments a seeded user account on each of those environments has been created. In addition a service for the user has been created as well as an email and sms template created.
 
 To run against those instances of Notify, additional environment variables for all of SMS_TEMPLATE_ID, EMAIL_TEMPLATE_ID and SERVICE_ID. These have already been set up on the Jenkins build using the settings page.
```
leohemsted, line 53: still references jenkins (and also line 58)

tests/functional/staging_and_prod/notify_api/test_notify_api_letter.py
idavidmcdonald: So my intention was that this would test that a precompiled letter uploaded makes it way through the antivirus app and then through template preview before successfully being marked as 'delivered' (although I realise here the test is checking 'received' and maybe I should change my test to check 'delivered').

If I look at logs for this test https://kibana.logit.io/s/a9805ba8-4219-49a6-9c9f-d466aceb842b/app/kibana#/discover?_g=(refreshInterval:(display:Off,pause:!f,value:0),time:(from:now-4h,mode:quick,to:now))&_a=(columns:!(message),index:'logstash-*',interval:auto,query:(query_string:(analyze_wildcard:!t,query:%22eb716a17-1cef-4997-ae4a-4d533387a4d2%22)),sort:!('@timestamp',desc))

You can see the initial API call to create it, follow by logs for antivirus, followed by the get request to check the status, followed by the template preview sanitisation, followed by the final get request. The fact the get requests only stopped after the letter had gone through template preview indicates to me that it was waiting for both antivirus and redaction to take place first, which is what i wanted to test.

If we update it to check for 'delivered' then it will check it makes it through to https://github.com/alphagov/notifications-api/blob/master/app/celery/letters_pdf_tasks.py#L290.

Note, this test is taken exactly from dev/preview and so I've taken the `NotificationStatuses.RECEIVED` from there too

idavidmcdonald: Ah, we want to keep it as `RECEIVED` because of https://github.com/alphagov/notifications-api/blob/master/app/models.py#L1572. So I think this test is correct as is and does test antivirus and template preview

leohemsted: fab, thanks for checking. the combination of the letter specific status aliases and test key stuff gives me a right headache

README.md
idavidmcdonald: Did a bit of a bigger tidy up. Jenkins is dead 💀

leohemsted: looks good

PR #328, opened 2020-04-20
Add init files for preview_dev and staging_prod projects

idavidmcdonald:
This should hopefully fix
https://github.com/pytest-dev/pytest/issues/3151

as we have seen the following error on concourse:
```
import file mismatch:
imported module 'test_notify_api_letter' has this __file__ attribute:
  /var/project/tests/functional/preview_and_dev/test_notify_api_letter.py
which is not the same as the test file we want to collect:
  /var/project/tests/functional/staging_and_prod/notify_api/test_notify_api_letter.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
```


idavidmcdonald:
This should fix the issue for the antivirus tests which are called in a unique way `{{ macros.run_functional_tests_task("preview", "-m antivirus") }}` rather than something like `{{ macros.run_functional_tests_task("preview", "tests/functional/preview_and_dev/") }}`

idavidmcdonald:
It won't fix the initial issue that Katie had with credentials by the way, just the issue with the antivirus func tests (and also template preview, if they were to run would also fall foul of this issue).

PR #329, opened 2020-04-29
Add a test for templated letters and remove research mode

klssmith:
### Stop using research mode service
We want to be able to test that templated letters get created, which is not possible in research mode, so we will turn off research mode for the functional test service.

### Add a test to see if templated letter is created
To see if a templated letter is created we check if a PDF file for it has been created in S3. It's not possible to do this by checking billable_units for the letter or something similar, because that does not get returned from notifications-api and we don't have access to the database to check.

To check if the letter is in S3 we use the client method of `get_pdf_for_letter`, which raises an exception if the letter is not found.

[Pivotal story](https://www.pivotaltracker.com/story/show/172413051)

**Not done**
I've left the name of the main functional test service as `RESEARCH_MODE...`. This could be changed, but it will need to be done in a few stages since we'd need to update credentials with the new name and it would involve devs changing their environment.sh files. I'd be keen to hear whether people think this is worth doing or not.

### DIFF
--- a/config.py
+++ b/config.py
@@ -134,8 +134,6 @@ def setup_staging_live_config():
 
         'notify_service_api_key': os.environ['NOTIFY_SERVICE_API_KEY'],
 
-        # this is either a live service or in research mode, depending on what tests are running
-        # (provider vs functional smoke tests)
         'service': {
             'id': os.environ['SERVICE_ID'],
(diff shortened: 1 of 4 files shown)


klssmith:
The tests pass on Concourse because I temporarily took the 'Functional Tests' service out of research mode while they ran

PR #330, opened 2020-05-01
Add email_access_validated_at data to setup SQL

tombye:
`email_access_validated_at` column was added to the API a while ago. This adds it to the setup SQL.


tombye:
> i'm almost surprised - is this the only change you needed to run this?

I've not run it against the whole suite but once `test_registration_and_invite.py` has been run others seem to run OK.

PR #347, opened 2020-08-28
Change how reply-to address id is found in page

tombye:
The ID for email reply-to addresses changed in
this pull request:

https://github.com/alphagov/notifications-admin/pull/3591

This updates the part of the test that finds the
id, to take in consideration the label added to
the front.


CrystalPea (approved):
👍

PR #348, opened 2020-09-01
Make xpath for team member link work with changes

tombye:
We changed the HTML for user-list-items in

https://github.com/alphagov/notifications-admin/pull/3605

This means the xpath that finds elements inside
the list items needs to work with this new HTML.

This changes the xpath to work with the existing
HTML and the new code.


servingUpAces (approved):

tombye:
The tests didn't pass first time because the new xpath didn't work for the existing HTML in the team members page. Once it found a matching `<h2>`, it backtracked too far before searching for a `<a>` tag, resulting in it returning more the wrong one.

We tested it locally but got a false positive because the link it returned worked for the test.

I've rewritten it to match an ancestor that exists in both versions of the HTML.

PR #367, opened 2021-02-03
Fix flakey / faulty tests for dev / preview

benthorner:
https://www.pivotaltracker.com/n/projects/1443052/stories/176742683

Please see the commits for more details.

### DIFF
--- a/config.py
+++ b/config.py
@@ -14,7 +14,7 @@ config = {
     # static
     'notification_retry_times': 15,
     'notification_retry_interval': 5,
-    'letter_retry_times': 48, 
+    'letter_retry_times': 48,
     'provider_retry_times': 12,
     'provider_retry_interval': 22,
     'verify_code_retry_times': 8,
--- a/tests/functional/preview_and_dev/test_seeded_user.py
+++ b/tests/functional/preview_and_dev/test_seeded_user.py
@@ -231,7 +231,7 @@ def test_view_precompiled_letter_message_log_delivered(
     )
 
     ref_link = config['service']['id'] + "/notification/" + api_integration_page.get_notification_id()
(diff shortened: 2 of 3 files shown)


benthorner:
Converting to "draft" as there's a discrepancy depending on whether these are run locally on or CI.

CrystalPea:
Hey, what do you mean by `log entry` in this context? I don't understand this and consequently do not understand this PR 😅

tombye:
I'm pretty sure `log_entry` refers to an entry in the 'Message log' list on `/services/<service_id>/api`. Ben has fixed some of my test code here so I recognise it 🙂

CrystalPea:
Ahh, sweet, thanks a lot @tombye

CrystalPea:
Ahh, I get it now @benthorner *galaxy_brain_moment*

benthorner:
Thanks for the question @CrystalPea. I assumed "everyone else will know what this is because they're all familiar with Notify" 😅. I had a bit of trouble choosing which words: in the UI it's described as a "message log" (see below), and in the code they're referred to as "notifications" ([here](https://github.com/alphagov/notifications-admin/blob/5bc229222be579ce21db26074f4e983fe9c12e74/app/templates/views/api/index.html#L38)), so I went with something in the middle.

Screenshot 2021-02-04 at 12 56 35

I've amended some of the commits and it's now passing in CI. Think it's ready for a review 👍.

benthorner:
After reviewing this myself, I realised the argument I'm making about the nature of the bug doesn't make sense. Rather than try and amend what's here, I've put up a fresh PR with a new branch/commits, and an explanation that really does match the evidence. See here: https://github.com/alphagov/notifications-functional-tests/pull/370.

PR #369, opened 2021-02-04
Test to see if CI behaves different to locally

benthorner:


benthorner:
For the record: the CI tests were passing (when they shouldn't).

PR #370, opened 2021-02-04
Fix test failing locally due to difference with CI

benthorner:
https://www.pivotaltracker.com/n/projects/1443052/stories/176742683

This was really hard to debug. Please see the second
commit for all the detail on the bug / fix.

## Example of the "API integration" page

Screenshot 2021-02-04 at 12 56 35

### DIFF
--- a/config.py
+++ b/config.py
@@ -14,7 +14,7 @@ config = {
     # static
     'notification_retry_times': 15,
     'notification_retry_interval': 5,
-    'letter_retry_times': 48, 
+    'letter_retry_times': 48,
     'provider_retry_times': 12,
     'provider_retry_interval': 22,
     'verify_code_retry_times': 8,
--- a/tests/functional/preview_and_dev/test_seeded_user.py
+++ b/tests/functional/preview_and_dev/test_seeded_user.py
@@ -489,6 +489,7 @@ def test_change_service_name(driver, login_seeded_user):
 
 def _check_status_of_notification(page, functional_tests_service_id, reference_to_check, status_to_check):
     page.go_to_api_integration_for_service(service_id=functional_tests_service_id)
+    page.expand_all_messages()
     client_reference = page.get_client_reference()
     assert reference_to_check == client_reference
     assert status_to_check == page.get_status_from_message()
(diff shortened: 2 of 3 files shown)


tombye:
Would it be worth having a helper method that looked for any open sections and closed them? It would mean you could reliably close anything open after actions, instead of keeping track of what is open. I was thinking of something like:
```python
open_buttons = driver.find_elements_by_css_selector('details[open] > summary')
for button in open_buttons:
    button.click()
```

benthorner:
> Would it be worth having a helper method that looked for any open sections and closed them? 

I did have a look at making the open/close behaviour more robust (see [first attempt PR](https://github.com/alphagov/notifications-functional-tests/pull/367)). Unfortunately this causes an issue with the "virus" test ([here](https://github.com/alphagov/notifications-functional-tests/blob/c94d05f39fc28d43dd82f23c9ae29257bb41a819/tests/functional/preview_and_dev/test_seeded_user.py#L315)), which (locally, where visibility matters) relies on having more than one entry open, as otherwise you'll get a "element not found" exception when it tries to look for a link that's not there. I thought about:

- Changing that test to expect "element not found", but that doesn't work on CI (because of the weirdness that it doesn't seem to care about visibility), which just picks the link in the second entry (even if it's closed).

- Adding a new "wait_for_no_element" helper in "pages.py", which would would use the "invisible" locator option (see [here](https://www.selenium.dev/selenium/docs/api/py/webdriver_support/selenium.webdriver.support.expected_conditions.html)) to check it's not there. But again, that would fail on CI, and just pick up the second log entry.

This is really what I meant in the commit when I said "simplest fix". Hopefully it's also fairly intuitive, too. Really we should investigate why CI is behaving differently, as that's not great, but would probably take even more time. For now, this seems like a net improvement: being slightly less code, but no more complex.

> The selenium debugging experience is pretty terrible, sorry.

Yes, it was a bit tricky to get into. Failing tests create a screenshot, though, which really helped (a lot!). Switching off headless would have been even better, though - I'll bear that in mind next time (hopefully there is no next time 😬).

leohemsted (approved):

PR #378, opened 2021-05-18
Add guidance for revalidating test email

benthorner:
This also fixes a flake8 issue, which was preventing the test script
from running locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,16 @@ pytest document_download # or
 pytest provider_delivery
 ```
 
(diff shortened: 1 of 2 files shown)


leohemsted (approved):
not something we need to do now

but i wonder if asserting that after entering the 2fa code we _don't_ get to a page that says "we've sent you an email address", then we'll fail on that page rather than the next one, and if you view the screenshot in concourse it'll be more obvious what needs to be done

benthorner:
> if you view the screenshot in concourse it'll be more obvious what needs to be done

I notice it prints out the file path of the screenshot, but I assumed it would be quite hard to get the file. 

Do you have a method?

PR #379, opened 2021-06-17
Rewrite and consolidate scripts

benthorner:
Contributes to: https://github.com/alphagov/notifications-manuals/issues/9

This continues the changes we've made in other repos to simplify the way we
run tests, as well as making the development environment more consistent.
Especially with this repo, the variety of ways to run the tests was confusing
to me, especially when the normal approach of running "pytest" just works.

Please see the commit messages for more details.


README.md
```diff
@@ -14,6 +14,8 @@ The Concourse jobs are defined in our [infrastructure repo](https://github.com/a
 
 ```shell
 brew install --cask chromedriver # needs to be >= v2.32
```
leohemsted, line 16: can we include a comment here about allowing chromedriver to run?

I had to go to:

System Preferences > Security & Privacy > General

where it says 

> "chromedriver" was blocked from use because it is not from an identified developer

and then i clicked "allow anyway"

you probably need to run the tests once to get the error before that appears there, i suspect.

leohemsted (approved):
couple of small comments but otherwise it works fine for me without using docker coming from a fresh setup

Makefile
```diff
@@ -1,168 +1,22 @@
...
 	@cat $(MAKEFILE_LIST) | grep -E '^[a-zA-Z_-]+:.*?## .*$$' | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
-.PHONY: venv
-venv: venv/bin/activate ## Create virtualenv if it does not exist
-
-venv/bin/activate:
-	test -d venv || virtualenv venv
+.PHONY: bootstrap
+bootstrap: ## Install build dependencies
+	pip install -r requirements.txt
```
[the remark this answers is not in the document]
benthorner: Good spot! I've fixed this in: 808107cbcc95d6324949ca7eceda1fc2a105b46e

README.md
benthorner: Fair point - I think I "just knew" when I set it up myself.

I've added a note about it in 808107cbcc95d6324949ca7eceda1fc2a105b46e

PR #382, opened 2021-07-07
Build Docker image using make task

benthorner:
Contributes to: https://github.com/alphagov/notifications-manuals/issues/9

This mirrors the approach we're taking in other repos [1][2], which
ensures the setup of the image is consistent with local setup.

[1]: https://github.com/alphagov/notifications-ruby-client/pull/122/commits/94858af5bbbe8ddf60ac9c544fcd96cb7f058cc3
[2]: https://github.com/alphagov/notifications-template-preview/pull/556

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+environment*.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,6 @@
 FROM joyzoursky/python-chromedriver:3.7-selenium
 
 WORKDIR /var/project
-ADD requirements.txt requirements.txt
-RUN pip install -r requirements.txt
-ADD tests/ tests/
-ADD config.py config.py
-ADD .flake8 .flake8
-ADD setup.cfg setup.cfg
-ADD pytest.ini pytest.ini
-ADD Makefile Makefile
-RUN mkdir logs
+COPY . .
+RUN make bootstrap
 ENTRYPOINT bash


idavidmcdonald (approved):
So reusing `make bootstrap` is sensible and good.

The bit that I want to point out as a very slight worry is the use of `COPY . .`. 

People may have files in the directory which are environment.sh files that they would be at risk of putting into docker images unintentionally. The risk is probably pretty low and maybe we should be solving the problem by removing the use of local environment files stored in the directory itself. But it still means secrets might get stored in places you weren't anticipating. It should be extremely rare for there to be a reason for a developer to have any production creds for running the functional tests, it's more often run against local or maybe preview.

I don't think it necessarily blocks this going in though unless it particularly concerns you

benthorner:
> environment.sh files that they would be at risk of putting into docker images unintentionally

It does seem very unlikely: we don't advocate Docker for local development with this repo, and CI builds images in a clean directory. If we did build the image locally, then yes there's a chance of secrets slipping in, and those could be production secrets if someone was developing tests that only run there.

I've amended the commit with a `.dockerignore` file to prevent secrets slipping in. That look OK to you?

I've also raised [a similar PR for the ruby-client](https://github.com/alphagov/notifications-ruby-client/pull/124), which is (currently) the only client with a `COPY . .` in its Dockerfile.

idavidmcdonald (approved):

PR #398, opened 2021-11-24
Update ‘Send’ button content

karlchillmaid:
We’re making this change because of this PR: https://github.com/alphagov/notifications-admin/pull/4075


quis (approved):
I think this is needed because `By.PARTIAL_LINK_TEXT` is case sensitive, and therefore ‘Send’ no longer finds ‘send’

PR #409, opened 2022-03-10
fix not being able to run broadcast tests locally

benthorner:
https://www.pivotaltracker.com/story/show/181498273

running the functional tests locally is a useful way to check my
local environment is working properly. these tests are also in the
"preview_and_dev" directory, so they should work locally.

this fixes the urls for local testing and also an inconsistency in
the service settings compared to preview: the tests check an alert
shows up on /alerts, but this won't happen for test alerts.

### DIFF
--- a/config.py
+++ b/config.py
@@ -48,11 +48,13 @@ urls = {
     },
     'staging': {
         'api': 'https://api.staging-notify.works',
-        'admin': 'https://www.staging-notify.works'
+        'admin': 'https://www.staging-notify.works',
+        'govuk_alerts': 'not used in this environment'
     },
     'live': {
         'api': 'https://api.notifications.service.gov.uk',
-        'admin': 'https://www.notifications.service.gov.uk'
+        'admin': 'https://www.notifications.service.gov.uk',
+        'govuk_alerts': 'not used in this environment'
     }
 }
 
--- a/config.py
+++ b/config.py
@@ -38,11 +38,13 @@ config = {
 urls = {
     'dev': {
         'api': 'http://localhost:6011',
-        'admin': 'http://localhost:6012'
+        'admin': 'http://localhost:6012',
+        'govuk_alerts': 'http://localhost:6017/alerts'
     },
     'preview': {
         'api': 'https://api.notify.works',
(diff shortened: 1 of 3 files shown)


config.py
```diff
@@ -38,11 +38,13 @@ def generate_unique_email(email, uuid):
...
         'api': 'http://localhost:6011',
-        'admin': 'http://localhost:6012'
+        'admin': 'http://localhost:6012',
+        'govuk_alerts': 'http://localhost:6017/alerts'
     },
     'preview': {
         'api': 'https://api.notify.works',
-        'admin': 'https://www.notify.works'
+        'admin': 'https://www.notify.works',
+        'govuk_alerts': 'https://www.integration.publishing.service.gov.uk/alerts'
```
idavidmcdonald, line 47: Optional: It might be worth adding a comment for the `staging` and `live` dicts to say why they don't have `govuk_alerts` keys? ie

```
'staging': {
        'api': 'https://api.staging-notify.works',
        'admin': 'https://www.staging-notify.works',
        # no govuk_alerts url for staging because... ?
    },
```

config.py
```diff
@@ -68,15 +70,14 @@ def setup_shared_config():
         'env': env,
         'notify_api_url': urls[env]['api'],
         'notify_admin_url': urls[env]['admin'],
+        'govuk_alerts_url': urls[env]['govuk_alerts']
```
idavidmcdonald, line 73: Will this cause an issue if any functional tests are run against staging or production because this line will `KeyError` as it does not have `govuk_alerts` as a key?

idavidmcdonald (commented):
I think there might be something that needs changing on this to do with config urls but do take a proper look and let me know if it might not actually be a problem!

config.py
benthorner: Great spot - thank you! Yes, we either need to make this conditional or use placeholder values. I've gone with the latter in https://github.com/alphagov/notifications-functional-tests/pull/409/commits/0d4dd8d9e739d0bf73e056df715ba2bb1946add5 as it's a convenient way to address your other comment.

idavidmcdonald (approved):
Lovely

PR #413, opened 2022-05-04
refresh web page if alert isn't present

leohemsted:
clicking the element by link text with page title ('Current alerts') won't work because that's not a link, it's just a h1. use driver.refresh instead.

Increase the timeout too

### DIFF
--- a/tests/pages/pages.py
+++ b/tests/pages/pages.py
@@ -1,6 +1,8 @@
 import os
 import shutil
+import time
 
+import pytest
 from retry import retry
 from selenium.common.exceptions import (
     NoSuchElementException,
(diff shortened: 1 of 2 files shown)


tests/pages/pages.py
```diff
@@ -1153,12 +1153,12 @@ def get(self):
         self.driver.get(self.gov_uk_alerts_url)
 
     def check_alert_is_published(self, page_title, broadcast_content):
```
benthorner, line 1155:
```suggestion
    def check_alert_is_published(self, broadcast_content):
```

Also, is there any way of testing the exception handler works as expected?

quis (approved):
A cursory glance suggests that maybe using the `@retry` decorator might have been cleaner, but this should work so let’s prioritise unblocking deploys.

PR #416, opened 2022-05-06
use a retry decorator to retry waiting for gov.uk alerts to render

leohemsted:
Use a retry decorator rather than a for loop, to be a bit more consistent and hopefully more readable

### DIFF
--- a/tests/pages/pages.py
+++ b/tests/pages/pages.py
@@ -1,8 +1,6 @@
 import os
 import shutil
-import time
 
-import pytest
 from retry import retry
 from selenium.common.exceptions import (
     NoSuchElementException,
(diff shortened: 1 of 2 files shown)


tests/pages/pages.py
```diff
@@ -1154,10 +1152,12 @@ def __init__(self, driver):
...
+    )
     def check_alert_is_published(self, broadcast_content):
-        for x in range(12):
-            if self.is_text_present_on_page(broadcast_content):
-                return
-            time.sleep(10)
+        if not self.is_text_present_on_page(broadcast_content):
             self.driver.refresh()
-        pytest.fail(f'Could not find alert with content "{broadcast_content}" despite waiting 2 mins')
+            raise RetryException
```
benthorner, line 1163:
```suggestion
            raise RetryException(f'Could not find alert with content "{broadcast_content}"')
```

We could still include this to get printed out as part of the final failure. I don't think it makes much difference, though, as someone will likely need to run manually to investigate anyway.

benthorner (approved):
Nice tidy up 👍.

PR #423, opened 2022-07-13
Refactor test user configuration 

rjbaker:
This change is in preparation for parallelising functional test suites.

Each test group generally uses a different user type. e.g. seeded_user, normal. broadcast_user etc. However when running the suite, all users are configured with the same mobile number.

When signing in a test user, the 2FA code is returned from the API by querying the mobile number. This causes a parallel user login session to be invalidated due to a race condition when querying the API for the 2FA code using the same number:

https://github.com/alphagov/notifications-functional-tests/blob/77a3254d28bc4d5b9e08302fa8517530742035ad/tests/test_utils.py#L279-L283

By using different numbers for the different user types, we can log each user type in simultaneously. This will allow us to run tests with different user types in parallel.

**Note:** This change does not run tests in parallel, so they will continue to safely run in sequence.

The following rows require updating to match the new mobile numbers added to the test configuration:

```sql
UPDATE users SET mobile_number='+447700900501' WHERE email_address='notify-tests-preview+admin_tests@digital.cabinet-office.gov.uk';
UPDATE users SET mobile_number='+447700900002' WHERE email_address='notify-tests-preview+broadcast1@digital.cabinet-office.gov.uk';
UPDATE users SET mobile_number='+447700900003' WHERE email_address='notify-tests-preview+broadcast2@digital.cabinet-office.gov.uk';
```

Paired with @sakisv


sakisv (approved):
Nice one 👍 

Related to alphagov/notifications-credentials#289

PR #424, opened 2022-07-14
Parallelise tests in preview environment

rjbaker:
### Context

Functional tests are slow because they run sequentially against a real environment. We can reduce test duration by running some tests in parallel.

### What does this do?

We have added the pytest [x-dist plugin](https://pypi.org/project/pytest-xdist/) to support running tests in parallel automatically. The number of test runners is determined automatically using the `pytest -n auto` option. This will be set to the number of CPUs available on the test machine.

Each test runner launches a separate selenium/chromedriver instance so browser state is isolated between runners.

We refactored test configuration in #423 to support simultaneous authentication of the different user types used in the test suite.

#### Defining parallel groups
We have chosen to execute tests in parallel using "groups" (`pytest --dist loadgroup`). This allows us to group tests by the authenticated user type or other logical domain - this is useful for functional tests that rely on a particular state of a real user environment during execution.

Parallel tests executed using the same user type can cause race conditions and test interference. Tests belonging to different groups are executed in parallel. Each test within a group is executed sequentially by the same test runner. 

We use the following annotations on test methods to define the groups:

```python
@pytest.mark.xdist_group(name="seeded-user")
@pytest.mark.xdist_group(name="registration-flow")
@pytest.mark.xdist_group(name="api-client")
@pytest.mark.xdist_group(name="seeded-email")
@pytest.mark.xdist_group(name="broadcasts")
@pytest.mark.xdist_group(name="api-letters")
```

A rough rule of thumb - More groups = better the parallelisation (limited by test runner count). However, in the case of functional tests, increased parallelisation increases the risk of side effects and race conditions in the shared environment unless grouped carefully.

### Results

Running sequentially, the "preview" functional test suite takes between `~14 - 17` minutes.
Running in parallel on a 4 CPU core Concourse worker, this is reduced to `~9` minutes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,5 +19,5 @@ clean: ## Remove temporary files
 test: clean ## Run functional tests against local environment
 	isort --check-only tests
 	flake8 .
-	pytest -v tests/functional/preview_and_dev
+	pytest -v tests/functional/preview_and_dev -n auto --dist loadgroup
 	pytest -v tests/document_download
--- a/README.md
+++ b/README.md
@@ -85,6 +85,33 @@ source environment_{env_name}.sh
 pytest tests/document_download
 ```
 
+### Running tests in parallel
+
+We can reduce the duration of the test suite by running some tests in parallel.
+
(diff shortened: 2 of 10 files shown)


sakisv (approved):

CrystalPea (approved):
Great work 👏🏼

Also I'm glad I could be of help 💖

PR #425, opened 2022-07-22
Update how the document download test calls document-download-api

klssmith:
document-download-api now accepts json content - it was still possible to send files while we switched over, but we now want to [only support json](https://github.com/alphagov/document-download-api/pull/205). This is the only place where we weren't already sending json to document-download-api.

### DIFF
--- a/tests/document_download/test_document_download.py
+++ b/tests/document_download/test_document_download.py
@@ -1,3 +1,5 @@
+import base64
+
 import pytest
 import requests
 from retry.api import retry_call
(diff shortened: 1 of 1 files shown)


tests/document_download/test_document_download.py
```diff
@@ -8,11 +8,11 @@
...
     response = requests.post(
-        "{}/services/{}/documents".format(config['document_download']['api_host'], service_id),
+        f"{config['document_download']['api_host']}/services/{service_id}/documents",
         headers={
-            'Authorization': "Bearer {}".format(config['document_download']['api_key']),
+            'Authorization': f"Bearer {config['document_download']['api_key']}",
         },
-        files={
+        json={
             'document': file_contents
```
leohemsted, line 16: should we base64 encode this?

klssmith: Yes, this should have been base64 encoded - I've now changed it.

leohemsted (approved):

PR #431, opened 2022-08-18
Change a locator for the Api integration page

klssmith:
We are changing the API integration page to use the GOV.UK Frontend details component. This means that it will no longer have a class called `api-notifications-item__heading` on the page, so we need to change one of the locators.

**Merge after the change to the API integration page** 
- [ ] https://github.com/alphagov/notifications-admin/pull/4322


tombye (approved):
Makes sense and looks good to me 👍🏻

PR #432, opened 2022-08-31
Revert "Update IDs of broadcast areas to use 2021 data set"

idavidmcdonald:
Reverts alphagov/notifications-functional-tests#428

PR #434, opened 2022-09-07
Revert "Revert "Update IDs of broadcast areas to use 2021 data set""

quis:
Reverts alphagov/notifications-functional-tests#432

Corresponds with https://github.com/alphagov/notifications-admin/pull/4343


samuelhwilliams (approved):
👍

PR #436, opened 2022-09-27
Delete some unused code

klssmith:


idavidmcdonald (approved):
Nice, any idea why this existed in the first place by chance? Did we use to have some functional tests that we then removed?

klssmith:
The deleted code all came from different commits where we'd refactored and forgotten to remove a function or class that had stopped being used. There were too many commits to list them all in the PR message, but to give a couple of examples:

`git log -p -S has_code_error` shows that we stopped using `has_code_error` in ca71b8a6ca9eeb73719b60d1243c2fd4b0ed2bbb
`git log -p -S click_upload_recipients` shows we stopped using `click_upload_recipients` in eb8527b13902b69a862a4a9b07d2075dec3f1a1f

PR #443, opened 2022-10-07
Add tests for doc dl email auth flow

leohemsted:
https://trello.com/c/rPd2HslH/16-add-functional-test-for-new-document-download-email-confirmation-flow

in the un-auth'd flow we use requests to just download the file and check the contents. If you use selenium to navigate chrome to that page, it'll just load an iframe with the pdf renderer inside so you cant assert if the content of the file is correct.

however, with the auth'd flow it relies on the cookie being set to access the file. requests doesn't have the cookie (only the chrome instances does) and i can't figure out how to get that cookie out, since driver.get_cookies() doesn't return it (it does return the other cookies on path '/').

so, to solve this, instead of getting a pdf, get a csv instead - when chrome renders that, it just shows the text with a minimal html wrapping, which is easy to assert on.

plus some other cleanup best looked at commit-by-commit

### DIFF
--- a/tests/document_download/test_document_download.py
+++ b/tests/document_download/test_document_download.py
@@ -4,21 +4,22 @@ from io import BytesIO
 import pytest
 import requests
 from notifications_python_client import prepare_upload
+from selenium.webdriver.common.by import By
 
 from config import config
-from tests.pages import DocumentDownloadLandingPage, DocumentDownloadPage
+from tests.pages import (
+    DocumentDownloadConfirmEmailPage,
+    DocumentDownloadLandingPage,
+    DocumentDownloadPage,
+)
 
 
-@pytest.mark.antivirus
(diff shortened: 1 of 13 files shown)


tests/document_download/test_document_download.py
```diff
@@ -73,10 +93,26 @@ def test_document_download_with_email_confirmation(driver, seeded_client):
...
+    else:
+        # chrome has downloaded the file
+
+        # get the filename out of the download URL
+        filename = urlparse(file_url).path.split("/")[-1]
+
+        document_path = get_downloaded_document(download_directory, filename)
+
+        with open(document_path) as f:
+            assert f.read() == "foo-bar-baz"
```
samuelhwilliams, line 115: Weird that we have to do this - but agree I can't work out why it's switching over and probably not worth investigating too much more.

Nicely handled 👍

PR #444, opened 2022-10-10
Split doc download func tests into preview vs staging and prod

leohemsted:
best reviewed commit by commit. did a lil cleanup while i was doing the following work.

## create separate preview and staging doc dl functional tests

we run smoke tests every ten minutes on prod - these can't use the seeded config as that is full of environment variables that are only used on preview on the full suite of tests.

So, follow the lead of hwo we've strucutred notify admin tests, and split into two folders - preview_and_dev and staging_and_live.

The preview_and_dev folder is just the old files unchanged.

The staging_and_live folder only has a single test to check a single document being sent. It uses the email confirmation flow just so we don't have to rewrite it later. (It's just the same test copy pasted).

we also remove the `@pytest.mark.antivirus` from the tests on the staging/prod file because that is a trigger to run the tests when antivirus merges, and should only apply to preview/staging methods.

----------------------

needs to be merged at the same time as:

- [ ] https://github.com/alphagov/notifications-aws/pull/1162


samuelhwilliams:
>we also remove the @pytest.mark.antivirus because that is a trigger to run the tests when antivirus merges, and should only apply to preview/staging methods.

I can only see an addition of one `pytest.mark.antivirus` - can't see any removed instances?

tests/conftest.py
```diff
@@ -97,23 +97,15 @@ def login_seeded_user(_driver):
...
+def client_live_key():
     client = NotificationsAPIClient(
         base_url=config["notify_api_url"], api_key=config["service"]["api_live_key"]
     )
     return client
 
 
 @pytest.fixture(scope="module")
-def seeded_client_using_test_key():
+def client_test_key():
```
samuelhwilliams, line 108: Great clean up 🙌

tests/document_download/staging_and_prod/test_document_download.py
```diff
@@ -0,0 +1,90 @@
...
+from config import config
+from tests.pages import (
+    DocumentDownloadConfirmEmailPage,
+    DocumentDownloadLandingPage,
+    DocumentDownloadPage,
+)
+from tests.test_utils import RetryException
+
+
+def _get_test_doc_dl_url(seeded_client, prepare_upload_kwargs):
```
samuelhwilliams, line 18: Think you copied these before renaming the clients so these don't resolve now, need to update to `client_live_key`

tests/document_download/staging_and_prod/test_document_download.py
```diff
@@ -0,0 +1,90 @@
...
+    file = prepare_upload(
+        BytesIO("foo-bar-baz".encode("utf-8")), **prepare_upload_kwargs
+    )
+    personalisation = {"build_id": file}
+    email_address = config["service"]["seeded_user"]["email"]
```
samuelhwilliams, line 23: Also `seeded_user` doesn't have an `email` key for staging+prod

tests/document_download/staging_and_prod/test_document_download.py
leohemsted: Good spot. updated and amended. I switched to `client_test_key` since we don't actually care about delivery at all, as doc download creates the link anyway

tests/document_download/staging_and_prod/test_document_download.py
leohemsted: Good catch, fixed. Tested against prod from my local. Tested against staging and I need to update the staging service to have doc dl contact details

samuelhwilliams: Cool, let me know when you've done that and I'll re-test on my end too

samuelhwilliams:
Tested this on preview myself now and looks good. Staging also looks good now except the doc dl service contact details.

tests/document_download/staging_and_prod/test_document_download.py
leohemsted: all good ```❯ source environment_staging.sh
❯ pytest tests/document_download/staging_and_prod
================================================== test session starts ===================================================
platform darwin -- Python 3.9.10, pytest-7.1.3, pluggy-1.0.0
rootdir: /Users/leo.hemsted/dev/functional-tests, configfile: pytest.ini
plugins: xdist-2.5.0, forked-1.4.0
collected 1 item

tests/document_download/staging_and_prod/test_document_download.py .                                               [100%]

=================================================== 1 passed in 8.77s ====================================================```

samuelhwilliams (approved):
What a pain 😬  Good work on this though - definitely leaving it in a better state than it was!

PR #449, opened 2022-11-07
find submit buttons by looking for a button inside a form

leohemsted:
they no longer have type=submit in govuk frontend v3.1.0 and higher (as that is implicit for any button without a type inside a form), so we need a different way to find them.

This is safe to merge before any changes to admin.

### DIFF
--- a/tests/pages/locators.py
+++ b/tests/pages/locators.py
@@ -61,7 +61,10 @@ class EditTemplatePageLocators(object):
 
 class UploadCsvLocators(object):
     FILE_INPUT = (By.ID, "file")
-    SEND_BUTTON = (By.CSS_SELECTOR, "main [type=submit]")
+    SEND_BUTTON = (
+        By.CSS_SELECTOR,
+        "form button.govuk-button:not(.govuk-button--secondary)",
+    )
     FIRST_NOTIFICATION_AFTER_UPLOAD = (By.CLASS_NAME, "table-row")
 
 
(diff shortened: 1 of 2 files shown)


tests/pages/locators.py
```diff
@@ -96,7 +96,7 @@ class InviteUserPageLocators(object):
         By.CSS_SELECTOR,
         "button[aria-controls=folder_permissions]",
     )
-    SEND_INVITATION_BUTTON = (By.CSS_SELECTOR, "main [type=submit]")
+    SEND_INVITATION_BUTTON = (By.CSS_SELECTOR, "form button")
```
tombye, line 99: When JS is running, there are 2 buttons on this page (assuming it is `/services/<service_id>/users/invite`):

image

Could do `form button.page-footer__button` instead to differentiate:

```suggestion
    SEND_INVITATION_BUTTON = (By.CSS_SELECTOR, "form button.page-footer__button")
```

tests/pages/pages.py
```diff
@@ -1107,7 +1107,7 @@ class ManageFolderPage(BasePage):
     delete_link = (By.LINK_TEXT, "Delete this folder")
     name_input = NameInputElement()
     delete_button = (By.NAME, "delete")
-    save_button = (By.CSS_SELECTOR, "main [type=submit]")
+    save_button = (By.CSS_SELECTOR, "form button")
```
tombye, line 1110: Same as last comment, with JS on this page has 2 buttons:

image

Same advice as before applies here:

```suggestion
    save_button = (By.CSS_SELECTOR, "form button.page-footer__button")
```

tombye (changes requested):
Spotted a few things but only minor. Once sorted this is good to go 👍🏻

PR #453, opened 2023-01-18
Move rename test to non seeded user

leohemsted:
if this test fails for whatever reason, it might leave the service in an unusual state (with the wrong service name). lots of tests check the service name to assert they're logged in to the correct account, and thus, will fail unless we manually rename the service back.

moving this test from the seeded service to the test registration service means we can not worry as much if the test fails, as no future test runs will look at this service.

this wasn't quite a lift-and-shift as i needed to tweak things because we don't know the service id, so now the change name test clicks the button in the service settings rather than just navigating to the URL - this is probably a better test anyway.

### DIFF
--- a/broadcast_client/broadcast_client.py
+++ b/broadcast_client/broadcast_client.py
@@ -1,7 +1,7 @@
 import urllib.parse
 
-from notifications_python_client.base import BaseAPIClient
 from notifications_python_client.authentication import create_jwt_token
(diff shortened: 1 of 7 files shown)


leohemsted:
last few func test preview runs:

463s, 459s, 478s

this func test PR run:

321s, 416s

samuelhwilliams (approved):

PR #459, opened 2023-04-03
Set `NOTIFY_ECS_ORIGIN` environment variable to force HTTP requests to go to ECS, not GOV.UK PaaS

idavidmcdonald:
https://trello.com/c/OzXPapgY/199-allow-configuration-of-functional-tests-to-include-http-header-to-divert-traffic-to-ecs

Best reviewed commit by commit

If you set the `NOTIFY_ECS_ORIGIN` environment variable, then because of our AWS Lambda function (https://github.com/alphagov/notifications-aws/tree/main/terraform/modules/migrate-traffic-to-ecs-lambda), you can make the functional tests run against our ECS infra, not our PaaS infra.

After we've migrated off the PaaS, we can remove all of this code.

To test this, start by checking you can run the functional tests against our preview PaaS environment from your local. You may need to first copy the credentials from `credentials/functional-tests/preview` into `environment_sh`. Then you can:

```
source environment_preview.sh
pytest tests/functional/preview_and_dev
```

These should go green.

Then run
```
NOTIFY_ECS_ORIGIN=true pytest tests/functional/preview_and_dev
```

Here you should see all failures. This is because our ECS infrastructure doesn't actually work yet... So failures are actually good because a failure means you are getting a change in behaviour when passing the environment variable...

If you want confidence that requests are indeed going to ECS, you can look at the metrics and monitoring for our ECS apps and see an uptick in memory and CPU when you run the functional tests, and also some errors in the logs too.

### DIFF
--- a/config.py
+++ b/config.py
@@ -50,13 +50,6 @@ urls = {
         "admin": "https://www.notify.works",
         "govuk_alerts": "https://www.integration.publishing.service.gov.uk/alerts",
     },
-    # ECS apps are being deployed piecemeal and at present there are no api or govuk_alerts
-    # urls for the ecs-preview environment, so for now we will use those of the preview env
-    "ecs-preview": {
-        "api": "https://api.notify.works",
-        "admin": "https://www-ecs.notify.works",
-        "govuk_alerts": "https://www.integration.publishing.service.gov.uk/alerts",
-    },
     "staging": {
         "api": "https://api.staging-notify.works",
         "admin": "https://www.staging-notify.works",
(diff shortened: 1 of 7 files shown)


sakisv (approved):
LGTM, but I'm not familiar with this repo so might be worth an extra pair of eyes 👀

sakisv (approved):
This looks good as far as I can tell.

Is it possible to replace the current `webdriver` module imported at line 6 and used at line 35? Is there a chance that these two things overlap and not work well together?

idavidmcdonald:
So I think this may answer your question...

The selenium webdriver documentation explicitely says that 'for sub-packages of webdriver, you should continue to import these directly from selenium'. Meaning that we should import both selenium and seleniumwire webdrivers

https://github.com/wkeeling/selenium-wire#creating-the-webdriver

sakisv:
I see, thanks 👍

PR #463, opened 2023-04-26
Add a functional test for CRUD letter template

samuelhwilliams:


tests/functional/preview_and_dev/test_seeded_user.py
```diff
@@ -143,6 +144,30 @@ def test_edit_and_delete_sms_template(driver, login_seeded_user, client_live_key
     assert current_templates == existing_templates
 
 
+@recordtime
+@pytest.mark.xdist_group(name="seeded-user")
```
leohemsted, line 148: does this need to be on the seeded user account? if we do this on the other account it will be a bit quicker (since the seeded users are currently the longest duration tests)

samuelhwilliams: mmm ... I don't know. But it's what the sms/email equivalent tests are on. 🤔

samuelhwilliams: So I reckon they don't strictly need the seeded user, but the other available user (email auth) doesn't have perms to manage templates. We could probably just give that user template permissions and move all three of these tests out of the seeded group?

samuelhwilliams: Provisionally added 12def0fa08e0953cdf2cbbe8b10ab84ccb8c50a6 - if we're happy with the email auth func test user having the 'manage templates' permission.

whpearson:
I think we'll need to update https://github.com/alphagov/notifications-functional-tests/blob/main/db_fixtures/local.sql to give the user permissions.

tests/functional/preview_and_dev/test_seeded_user.py
samuelhwilliams: Ok I don't think we can seamlessly just send these ones over to the email auth user, because those run in parallel and have caused some conflicts (that user gets a password changed/reset, these tests try to log in with a static password, caused a failure like this one: https://concourse.notify.tools/teams/notify/pipelines/apps/jobs/functional-tests-preview/builds/631)

samuelhwilliams:
Good spot @whpearson, we probably would/will have to. For now I'm not going to change those perms, but good to call it out if/when we do

tests/pages/pages.py
```diff
@@ -571,6 +575,43 @@ def create_template(self, name="Test sms template", content=None):
...
 
+class ViewLetterTemplatePage(BasePage):
+    edit_body = ViewLetterTemplatePageLocators.EDIT_BODY
+
+    def click_edit_body(self):
+        element = self.wait_for_element(ViewLetterTemplatePage.edit_body)
+        element.click()
+
+
+class EditLetterTemplatePage(BasePage):
```
whpearson, line 586: As this is similar to the EditSMSTemplatePage is it worth creating a base class?

samuelhwilliams: I've updated it to reuse EditSmsTemplatePage 👍

tests/pages/pages.py
```diff
@@ -571,6 +575,28 @@ def create_template(self, name="Test sms template", content=None):
...
+    def click_edit_body(self):
+        element = self.wait_for_element(ViewLetterTemplatePage.edit_body)
+        element.click()
+
+
+class EditLetterTemplatePage(EditSmsTemplatePage):
+
+    name_input = NameInputElement(clear=True)
+    template_content_input = TemplateContentElement(clear=True)
+
```
BlessedDev, line 590: This couples `EditLetterTemplatePage` to `EditSmsTemplatePage` in a way that might be in a nuisance in future, if we for instance have to make changes to `EditSmsTemplatePage` that we don't want to carry over to `EditLetterTemplatePage`. 
Given the similarities between `EditLetterTemplatePage`, `EditSmsTemplatePage` and `EditEmailTemplatePage` I agree with Will that a base class would be worth it.

BlessedDev (commented):
Looks good to me, I have left a comment for a minor issue for you to consider.

tests/pages/pages.py
samuelhwilliams: I don't care much for a base class really. They are different pages and if we have some apparent duplication I don't think it costs us anything significant.

I made an update after @whpearson commented to follow what the existing `EditBroadcastTemplatePage` does and just pretend to be a sub-class of `EditSmsTemplatePage`, but on reconsideration I'm going to move in the opposite direction and just go back to full duplication.

I don't think inheritance here gives us as much value as we think.

BlessedDev (approved):

PR #470, opened 2023-06-09
Wp func test add remove attachment

whpearson:
It

- adds a new template
- attaches a blank page to that template
- checks that the attach pages button is now manage pages
- checks the filename on the manage page is present
- deletes the attachment
- deletes the template


tests/functional/preview_and_dev/test_seeded_user.py
```diff
@@ -168,6 +171,20 @@ def test_edit_and_delete_letter_template(driver, login_seeded_user, client_live_
     assert current_templates == existing_templates
 
 
+@recordtime
+@pytest.mark.xdist_group(name="seeded-user")
+def test_add_and_delete_letter_attachment(driver, login_seeded_user, client_live_key):
+    test_name = "edit/delete letter attachment test"
+    go_to_templates_page(driver)
+    create_letter_template(driver, name=test_name, content=None)
+    add_letter_attachment_for_template(driver, name=test_name)
```
samuelhwilliams, line 180: In my testing of this we are consistently hitting the timeout exception, which will cause the test to run a lot slower than it needs to. I think we need to navigate back to the "list templates" page before doing this.

Same with the "delete_template" step at the end - let's explicitly navigate back to list templates first.

Not sure if you did, but it can be helpful to run the test with `--no-headless` and watch it run through. It becomes fairly obvious when the browser isn't doing anything, which suggests a place where we're looking for an element that doesn't exist. Avoiding those would be great 👌 

Otherwise I think this test looks good and nice work for adding it

whpearson: Nice, it is 20s quicker now (with 1 statistically insignificant test)

samuelhwilliams (approved):
Nice, this feels much faster now 👍

PR #471, opened 2023-06-12
Add tag to tests that should pass locally

samuelhwilliams:


samuelhwilliams:
I feel like `test_view_precompiled_letter_message_log_virus_scan_failed` should be able to pass locally but I can't make it yet 🤔  And only throwing this up, not gonna dig deeper right now...

PR #473, opened 2023-08-17
Add functional test validating QR code data

samuelhwilliams:
This creates a test which sends a letter with a QR code in it. We then grab the PDF for the letter and look for a QR code in it, then check that the QR code data matches what we expected.


whpearson (approved):
Changes look good

PR #474, opened 2023-08-18
Seeded user per test

samuelhwilliams:
(Requires https://github.com/alphagov/notifications-credentials/pull/377 and https://github.com/alphagov/notifications-api/pull/3851 to be merged first for CI tests to pass)

---

The slowest group of tests in our functional test suite at the moment
is the xdist group 'seeded-user'. These tests all run using a
pre-defined user account in a specific organisation and service, with
specific permissions. The tests don't necessarily step on each other's
toes that much, but our app enforces only one user session can be active
at a time. If two of the seeded-user tests run at the same time, one of
the sessions will be killed by the other.

To deal with this, let's instead use a new private API endpoint to
create a 'seeded' user for each test that needs one. This will let all
of those tests run in parallel and give us a much faster runtime for the
full suite of tests.

At startup, we analyse all of the tests that have the
'login_seeded_user' fixture, and post a request to the API endpoint to
make sure that users exist for all of those tests. We then rewrite the
`get_email_and_password` helper to return the unique per-test email
address. We only execute this flow on one of the workers using a
temporary file to guarantee only a single execution.

There are a few tests that read/write templates and other central
objects at the same time. Where tests do that, we make sure to either
create each template with a unique name, or use an arbitrary locking
method to ensure only one test can be inside a given block of code at a
time. The latter is most relevant for letter template tests, as letter
templates are all created with the name 'Untitled letter template' and
need to be made unique in multiple steps.

For a locking mechanism we can use a simple file-based lock, as the test
run is not executed across multiple machines.

---

I'm hopeful we can extend this de-grouping effort to other xdist groups at some point in the future, with the ideal situation (probably) being all tests being able to run independently (and ideally removing the concept of "pre-existing"/seeded DB fixtures altogether)

# Timing

I've excluded the broadcast tests as these should be disappearing from our suite soon. And at the time of writing these were running slow ([probably related to this](https://github.com/alphagov/notifications-manuals/wiki/Useful-queries#delete-old-broadcast-messages))

## Before - 6m46
Notice that even though we spin up 10 workers, we can only use 5 because we only have 5 xdist loadgroups.
image

## After - 2m13s
Notice that we can now actually use all of our workers, and so do more work at once.
image

<details><summary>Taking it one step further...</summary>
<p>
We have the possibility of taking this even further by increasing the number of workers beyond that provided by `-nauto` (8 on concourse). Because a fair chunk of the work is waiting for IO across the network, we can probably have more workers than CPUs.

By bumping workers up to 16 locally (vs preview) I managed to get runtime down to 1m44. Would need to validate whether this is worth doing on concourse though.

image
</p>
</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,6 @@ Parallel tests executed using the same user type can cause race conditions and i
 We use the following annotations on test methods to define the groups:
 
 ```python
-@pytest.mark.xdist_group(name="seeded-user")
 @pytest.mark.xdist_group(name="registration-flow")
 @pytest.mark.xdist_group(name="api-client")
 @pytest.mark.xdist_group(name="seeded-email")
--- a/config.py
+++ b/config.py
@@ -94,6 +94,10 @@ def setup_preview_dev_config():
                 "password": os.environ["FUNCTIONAL_TEST_PASSWORD"],
                 "mobile": os.environ["TEST_NUMBER"],
             },
+            "api_auth": {
+                "client_id": "notify-functional-tests",
+                "secret": os.environ["FUNCTIONAL_TESTS_API_AUTH_SECRET"],
+            },
             "notify_service_api_key": os.environ["NOTIFY_SERVICE_API_KEY"],
             "broadcast_service": {
                 "id": os.environ["BROADCAST_SERVICE_ID"],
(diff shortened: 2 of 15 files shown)


CrystalPea (approved):
Great work making integration tests so much faster 👏🏼 . 

[note for the future] Quite a few changes were squished into one commit - it would have been easier to review and see various decisions if it was a few commits instead.

PR #475, opened 2023-08-20
Bump letter allowance to 1k locally

samuelhwilliams:
There's no real reason to have such a restrictive letter sending limit for local DB fixtures. All it means is that if you run the functional tests a few times on your local machine you'll start getting weird unexpected errors.

Let's just make it 1k to be more generous.


idavidmcdonald (approved):
Nice.

I'm sure you are aware that this won't take effect on any devs local machines unless (I think) they nuke their local database and then rerun this script. But that is OK and a limitation of our test fixtures.

samuelhwilliams:
Yep of course :) But at least stops the problem occurring for new devs.

PR #476, opened 2023-08-21
Add pytest-rerunfailures

samuelhwilliams:
We have some occasionally tests that flake due to selenium not being able to find an element on the page. In CICD as this stands, it will fail the suite and cause the entire suite to be re-run in a new container. This is an undue delay if it is truly a flaky issue.

This plugin will let pytest automatically try to re-run the test a few times (configurable). Assuming it is a flake, rerunning just that one failed test will save a lot of time over having to go through the whole suite again.

Will need a corresponding change on the pipeline to actually enable it.


leohemsted:
i'm a little concerned about the actual repeatability and sequencing of tests (especially if something fails the state of a test involving the seeded service could be an issue) but we can keep an eye on this

PR #479, opened 2023-08-31
Disable broadcast tests locally

samuelhwilliams:
Most of us don't run gov.uk alerts locally, and we will soon be stopping to support the EA code. Let's exclude broadcast functional tests from running locally by default.


environment_local.sh
```diff
@@ -36,3 +36,4 @@ export BROADCAST_USER_2_NUMBER=07700900003
 
 export BROADCAST_SERVICE_ID='8e1d56fa-12a8-4d00-bed2-db47180bed0a'
 export BROADCAST_SERVICE_LIVE_API_KEY='func_tests_broadcast_service_live_key-8e1d56fa-12a8-4d00-bed2-db47180bed0a-c3e6b68f-bd43-4e33-8fba-67a8c1bad4a3'
+export FUNCTIONAL_TESTS_EXTRA_PYTEST_ARGS='-k "not broadcast"'
```
klssmith, line 39: We've got a Pytest marker for the broadcast tests, so you could use `-m 'not broadcast'` if you wanted (which is what we do for the ECS tests) but either is fine

klssmith (approved):

environment_local.sh
samuelhwilliams: That's a bit more exact, so good suggestion 👍

PR #480, opened 2023-09-18
Update letter attachment functional test

samuelhwilliams:
Adds a step to the "add and remove letter attachment" test which also sends a letter with an attachment and verifies the attachment is present in the rendered PDF.

Not split out into a separate test because it would by definition end up duplicating most of the steps anyway (adding an attachment)


tests/functional/preview_and_dev/test_seeded_user.py
```diff
@@ -162,15 +164,49 @@ def test_edit_and_delete_letter_template(driver, login_seeded_user, client_live_
...
     add_letter_attachment_for_template(driver, name=template_name)
     assert driver.find_element(By.CLASS_NAME, "edit-template-link-attachment").text == "Manage attachment"
+
+    send_letter_to_one_recipient(
+        driver, template_name, address=f"{test_id}\nTest street\nSW1 1AA", build_id=str(test_id)
+    )
+
+    letter_preview_page = PreviewLetterPage(driver)
+    pdf_download_link = letter_preview_page.get_download_pdf_link()
+    notification_id = pdf_download_link.split("/")[-1].replace(".pdf", "")
```
CrystalPea, line 182: [suggestion] we could use existing method for getting a notification ID for simplicity ☺️ : https://github.com/alphagov/notifications-functional-tests/blob/44770b4c1ec484c90bcf1048b3bf6290a5c4c0e1/tests/pages/pages.py#L413

tests/functional/preview_and_dev/test_seeded_user.py
```diff
@@ -162,15 +164,49 @@ def test_edit_and_delete_letter_template(driver, login_seeded_user, client_live_
...
+
+    retry_call(
```
CrystalPea, line 184: [suggestion] since we don't have an explicit assertion here, a one-line comment explaining what we do here would aid readability, for example:

`# checking that the letter was created`

tests/functional/preview_and_dev/test_seeded_user.py
```diff
@@ -162,15 +164,49 @@ def test_edit_and_delete_letter_template(driver, login_seeded_user, client_live_
...
+        ],
+        tries=config["notification_retry_times"],
+        delay=config["notification_retry_interval"],
+    )
+
+    letter_pdf = get_pdf_for_letter_via_api(client_live_key, notification_id)
+    pdf_reader = PdfReader(letter_pdf)
+    assert len(pdf_reader.pages) == 2
+    attachment_page = pdf_reader.pages[1]
+    assert "This is an attachment" in attachment_page.extract_text()
```
CrystalPea, line 199: Nice tight assertion 👏🏼

CrystalPea (approved):
Nice one 👍🏼 . A couple of suggestions below:

PR #484, opened 2023-10-04
Testing edit sms/email templates

samuelhwilliams:


leohemsted (approved):
the "test_end_and_delete_email_template" test doesn't actually test editing :sob:

samuelhwilliams:
now it does 🎉  but yes 😢

PR #485, opened 2023-10-04
handle if we're already signed in when trying to sign in

leohemsted:
many tests sign in at the beginning. if they then fail, they'll be retried on concourse but the user will stil be signed in. the first thing the second run of the test does is try and sign in again, which will throw an exception when it tries to go to the sign_in page cos it'll redirect to /accounts-and-services which the functional tests don't expect.

so instead, catch this error, and try to sign out before continuing

### DIFF
--- a/tests/pages/rollups.py
+++ b/tests/pages/rollups.py
@@ -29,7 +29,7 @@ def _sign_in(driver, account_type, test_name=None):
     sign_in_page.get()
 
     try:
-        sign_in_page.is_current()
+        sign_in_page.wait_until_current()
     except TimeoutException:
         # if we didn't get to the sign_in_page, it's probably because we're already logged in.
         # try logging out before proceeding
--- a/tests/pages/rollups.py
+++ b/tests/pages/rollups.py
@@ -3,6 +3,7 @@ import os
 import tempfile
 
 from filelock import FileLock
+from selenium.common.exceptions import TimeoutException
 
 from config import config, generate_unique_email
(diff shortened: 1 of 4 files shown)


samuelhwilliams:
What test is this an issue with? I haven't seen this happening and I've seen seeded user tests rerun and then pass quite a few times 🤔

leohemsted:
so i haven't checked, but i suspect seeded user tests work fine because the sign_in is in the _fixture_

this issue was observed with the broadcast tests but could happen with any test that calls sign_in manually

tests/pages/rollups.py
```diff
@@ -26,7 +27,15 @@ def sign_in_email_auth(driver):
 def _sign_in(driver, account_type, test_name=None):
     sign_in_page = SignInPage(driver)
     sign_in_page.get()
-    assert sign_in_page.is_current()
+
+    try:
+        sign_in_page.is_current()
```
samuelhwilliams, line 32: meh this reads weird, looks like it's just a test returning a bool but it's a wait-until instead.

even more weirdly _some_ of the `is_current` calls are just quick comparisons and some are wait-untils, super inconsistent 🙃

samuelhwilliams (approved):

tests/pages/rollups.py
leohemsted: yeah you're right. i'm gonna rename this to `wait_until_current`

samuelhwilliams (approved):

PR #493, opened 2024-01-08
Python 3.11

leohemsted:
This was a bit wilder than expected - the dockerfile we normally use that provides both a specific python version and chromedriver isn't available on dockerhub for python 3.11 - but the dockerfile is there in the source github repo. So we took that dockerfile and duplicate it.

I then needed to make a bunch of changes to it in 6fb56ec88b96d2e3614ecfa458641f864e81a20e: description as follows.

----

[install chromedriver that matches the version of chrome](https://github.com/alphagov/notifications-functional-tests/commit/6fb56ec88b96d2e3614ecfa458641f864e81a20e) 

google-chrome-stable currently returns 120.0.6099.199. however, for reasons i can't figure out, the chromedriver website[^1] only tops out at version 115, otherwise it points you to the chromedriver for testing page. (Note that 115 is six months old. 120 is only a week or so old). You need to have a matching version of google-chrome and chromedriver, so we need to either test against an old version of chrome or make sure we get a newer version of chromedriver. Pinning to an old version seems likely to fall behind pretty quick, so instead get the latest version of chromedriver as referred to in last-known-good-versions.json on the chrome github[^2].

The eagle eyed may notice that there's still a difference between the stable as reported by googlechromelabs.github.io and the stable as returned in the debian package (120.0.6099.199 vs .109). The correct answer is probably to also download google chrome from source from the github link (so we know we have a totally matching version). However, for now while these two work I'm okay to leave it as is to avoid us making this dockerfile even more handwritten.

Also note that the unzip had to change slightly as the zip file layout changed from the old chromedriver website to the new github repo link

[^1]: https://chromedriver.chromium.org/downloads
[^2]: https://github.com/GoogleChromeLabs/chrome-for-testing#json-api-endpoints

----------

i've tested this locally by building the docker image and running functional tests against preview. It might need a little more testing.

We shouldn't need to update any concourse - rather we'll need to merge this, wait for it to build the image, and then see what breaks 💥.


risicle:
:see_no_evil: the unreproducibility :see_no_evil: 

Guess it's not really any worse than what upstream were already doing behind the scenes though.

CrystalPea (approved):
wow, what a journey 😳

PR #499, opened 2024-03-14
Update tests now that they only need to be run from one place (ECS)

klssmith:
This deletes all the code that was added to let the tests run on both PaaS and ECS.


idavidmcdonald (approved):
We have now stopped running the `functional-tests-production` jobs ie functional tests now only run against ECS. So this code looks good, and I think can now be merged!

PR #503, opened 2024-04-24
Update labels for editing a letter template

quis:
To match https://github.com/alphagov/notifications-admin/pull/5021


tombye (approved):
Makes sense to me 👍🏻

PR #505, opened 2024-05-20
Update CHANGE_SERVICE_NAME_LINK selector

kr8n3r:
In https://github.com/alphagov/notifications-admin/pull/5078 we updated the markup so CHANGE_SERVICE_NAME_LINK needs to be updated as well


leohemsted (approved):

tombye (approved):
Looks good to me and the selector works when tested locally 👍🏻

PR #508, opened 2024-06-06
Adding script to process sender names from BT file

whpearson:
See readme for details of how to use it.


idavidmcdonald:
Just a quick drive by comment. I wonder if this script would be better in the notifications-api repo. It feels like seeding information into the database (if not related to functional test fixtures) would best live in the repo that is closest to our database which I think is notifications-api.

Thoughts?

notifications-aws would also be a contender but I think I lean towards notifications-api

Open to thoughts!

whpearson:
Happy to put it wherever. I'll add it to the api repo

PR #509, opened 2024-06-11
Add test for inbound SMS callbacks

robinjam:
[Trello card](https://trello.com/c/4i0nHh9o/760-functional-tests-for-sms-inbound-callbacks)

This test relies on a third party request bin service[^1]. We configure the Functional Tests service on Notify to send inbound SMS callbacks to a source configured on Pipedream. We can then use the Pipedream API to verify that the correct callback was triggered when an inbound SMS was received.

[^1]: https://pipedream.com


tests/functional/preview_and_dev/test_inbound_sms.py
```diff
@@ -53,6 +54,21 @@ def test_inbound_api(inbound_sms, client_live_key):
     next(x for x in client_live_key.get_received_texts()["received_text_messages"] if x["content"] == inbound_sms)
 
 
+def test_inbound_sms_callbacks(inbound_sms):
+    sleep(0.2)  # Allow time for the callback to be sent
```
corlettb, line 58: This sleep looks like something that will let us down often. Sleep is not used anywhere else in the code.

I think the approach the tests tend to use is 'retry_call'. E.g. https://github.com/alphagov/notifications-functional-tests/blob/main/tests/functional/preview_and_dev/test_notify_api_letter.py#L66

risicle (approved):
Turned out smaller and neater than I'd dare hope.

Guess we can't do much more than suck it & see.

corlettb (approved):

PR #515, opened 2024-08-16
check status code for pipedream

leohemsted:
it's currently failing because request['body'] isn't present - i've got a suspicion that the request might be erroring - this is probably a first step to building slightly better more robust debugging statements in to this


robinjam (approved):

robinjam:
@leohemsted I think this is still worth merging even if the tests are green now - would you be able to fix the ruff error that's blocking the merge or do you want me to do it?

leohemsted:
oops good spot @robinjam, i missed that cos I just edited in the github editor

if you re-✅ i can merge

robinjam (approved):

PR #517, opened 2024-08-28
Use Debian-packaged chromium & chromium-driver

robinjam:
[Trello card](https://trello.com/c/9AWH7VCa/875-use-chromium-and-chromium-driver-from-debian-apt-repo-in-builds-of-the-functional-test-docker-image)

We currently install the latest version of Chrome from Google’s apt repo, and the latest version of Chromedriver from chrome-for-testing. This is brittle, because it’s not guaranteed that we’ll end up with matching versions of Chrome and Chromedriver.

We should use Debian’s own chromium and chromium-driver packages instead.

Selenium seems to be able to find Chromium automatically (even though it lives in a different path than Chrome), so we don't need to make any changes to our test config.

We will continue pinning Chromium v126 for now, because v127 and later are incompatible with our functional tests. This will need to be followed up with some more work to investigate why v127+ isn't working (so that we can unpin v126), and/or switch to a different browser such as Firefox.

Tested [with the existing CI for this repo](https://concourse.notify.tools/teams/notify/pipelines/apps/jobs/build-functional-tests-pr/builds/286) and [on dev-c](https://concourse.notify.tools/teams/dev-c/pipelines/deploy-notify/jobs/test/builds/18).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,48 +1,24 @@
-# The jorzoursky/python-chromedriver image does not have a published 3.11 tag yet, so copy the dockerfile across from
-# https://github.com/joyzoursky/docker-python-chromedriver/blob/master/py-debian/3.11-selenium/Dockerfile
-FROM python:3.11 as python-chromedriver-copy
+FROM python:3.11
 
-# install google chrome
-RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add -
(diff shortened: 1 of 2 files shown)


robinjam:
As for why Selenium is able to find Chromium without needing to be configured: I suspect it's because the `chromium-driver` package provides `/usr/bin/chromedriver`: https://packages.debian.org/bookworm/amd64/chromium-driver/filelist

risicle (approved):
Let's not forget about this for too long.

PR #523, opened 2024-10-09
Increase timeout before failing letter sending smoke test

robinjam:
Increases the timeout from 4 minutes (48 retries with 5 seconds between them) to 9 minutes (108 retries with 5 seconds between them).

Letter sending is less time-sensitive than other types of notifications, so we don't want to be alerted unless the delay is significant. The smoke tests run every 10 minutes, so we don't want each run of the test to take longer than that.


tombye (approved):
Looks good to me 👍🏻

PR #524, opened 2024-10-18
Add unsubscribe request functional tests

BlessedDev:
This PR adds a functional test that covers the entire email unsubscribe request journey as shown in [this ](https://trello.com/c/7ngdhTeg/80-add-a-functional-test-for-the-manual-email-unsubscribe-process)card.

The only thing not addressed explicitly in this PR is the cleanup of the unsubscribe requests generated. Instead this PR is relying on the fact that the unsubscribe requests generated will be archived after 7 days, https://github.com/alphagov/notifications-api/blob/a648d502cfc9d782145622b54f25f8262c1c963e/app/celery/nightly_tasks.py#L79

Each time the test runs, two unsubscribe requests are generated one for each pathway an unsubscribe request is generated ie. either via the the link in the email header or in the body of the email.

### DIFF
--- a/tests/functional/preview_and_dev/test_unsubscribe_requests.py
+++ b/tests/functional/preview_and_dev/test_unsubscribe_requests.py
@@ -24,6 +24,7 @@ def test_unsubscribe_request_flow(request, driver, login_seeded_user, client_liv
 
     dashboard_page = DashboardPage(driver)
     dashboard_page.go_to_dashboard_for_service(service_id=config["service"]["id"])
+    dashboard_email_unsubscribe_stats_before = dashboard_page.get_email_unsubscribe_requests_count()
 
     # Send the notification via api
     send_notification_to_one_recipient(
@@ -56,6 +57,8 @@ def test_unsubscribe_request_flow(request, driver, login_seeded_user, client_liv
 
     # Go to Email unsubscribe requests summary page
     dashboard_page.go_to_dashboard_for_service(service_id=config["service"]["id"])
+    dashboard_email_unsubscribe_stats_after = dashboard_page.get_email_unsubscribe_requests_count()
+    assert dashboard_email_unsubscribe_stats_after > dashboard_email_unsubscribe_stats_before
     dashboard_page.click_email_unsubscribe_requests()
 
     # Go to email unsubscribe request report page
--- a/tests/functional/preview_and_dev/test_unsubscribe_requests.py
+++ b/tests/functional/preview_and_dev/test_unsubscribe_requests.py
@@ -1,0 +1,35 @@
+import os
+import uuid
+
+from config import config
+from tests.pages import DashboardPage
+from tests.test_utils import create_email_template, go_to_templates_page, recordtime, send_notification_to_one_recipient
+
+
+@recordtime
+def test_unsubscribe_request_flow(request, driver, login_seeded_user, client_live_key):
+    # Create a subscription template
+    go_to_templates_page(driver)
+    template_name = f"functional subscription email template {uuid.uuid4()}"
+    content = "Hi ((name)), Is ((email address)) your email address? We want to send you some ((things))"
+    template_id = create_email_template(driver, name=template_name, content=content, has_unsubscribe_link=True)
+
+    dashboard_page = DashboardPage(driver)
+    dashboard_page.go_to_dashboard_for_service(service_id=config["service"]["id"])
+
+    # Send the notification via api
+    send_notification_to_one_recipient(
+        driver,
+        template_name,
+        "email",
+        test=False,
+        recipient_data=os.environ["FUNCTIONAL_TEST_EMAIL"],
+        placeholders_number=2,
+    )
+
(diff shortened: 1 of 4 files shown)


tests/functional/preview_and_dev/test_unsubscribe_requests.py
```diff
@@ -0,0 +1,71 @@
...
+    path = urlparse(generated_one_click_unsubscribe_url).path
+    admin_url = urljoin(urls[os.environ["ENVIRONMENT"]]["admin"], path)
+    driver.get(admin_url)
+
+    unsubscribe_request_confirmation_page = UnsubscribeRequestConfirmationPage(driver)
+    unsubscribe_request_confirmation_page.click_confirm()
+    assert urlparse(driver.current_url).path == "/unsubscribe/confirmed"
+
+    # Go to Email unsubscribe requests summary page
+    dashboard_page.go_to_dashboard_for_service(service_id=config["service"]["id"])
```
CrystalPea, line 56: Do we want to check the count of unsubscribe requests on the dashboard? I remember we wanted to check it, but run into some issues?

tests/pages/pages.py
```diff
@@ -975,6 +993,10 @@ def send_to_myself(self, message_type):
             element = self.wait_for_element(SingleRecipientLocators.USE_MY_NUMBER)
         element.click()
 
+    def click_send_one_email_button(self):
```
CrystalPea, line 996: I don't see us using this function anywhere? If we are not using it, could we remove it, and `SEND_ONE_EMAIL_BUTTON`?

tests/pages/pages.py
```diff
@@ -1211,3 +1233,44 @@ def delete_attachment(self):
         delete_button.click()
         confirm_button = self.wait_for_element(self.confirm_button)
         confirm_button.click()
+
+
+class UnsubscribeRequestConfirmationPage(BasePage):
+    confirm_unsubscription_button = (By.CSS_SELECTOR, "button[type=submit]")
+
+    def click_confirm(self):
```
CrystalPea, line 1241: Would one of BasePage methods, like `click_save()` or `click_continue()` work here? They should if it's just a standard button, and then we could reduce the amount of new boilerplate?

tests/pages/pages.py
```diff
@@ -1211,3 +1233,44 @@ def delete_attachment(self):
...
+    unsubscribe_request_report_link = (By.CSS_SELECTOR, "th a")
+
+    def click_latest_unsubscribe_request_report_by_link(self):
+        element = self.wait_for_element(UnsubscribeRequestReportsSummaryPage.unsubscribe_request_report_link)
+        element.click()
+
+
+class UnsubscribeRequestReportPage(BasePage):
+    download_report_link = (By.LINK_TEXT, "Download the report")
+    back_link = (By.LINK_TEXT, "Back")
```
CrystalPea, line 1256: a back link is quite a generic selector, maybe we want to put it in `class CommonPageLocators`, or somewhere like that?

tests/functional/preview_and_dev/test_unsubscribe_requests.py
```diff
@@ -0,0 +1,71 @@
...
+
+    # Download email unsubscribe request report
+    report_page = UnsubscribeRequestReportPage(driver)
+    report_page.click_download_report_link()
+    report_page.click_back_link()
+
+    # Mark a report as completed
+    unsubscribe_request_reports_summary_page.click_latest_unsubscribe_request_report_by_link()
+    report_page.select_mark_as_complete_checkbox()
+    report_page.click_update_button()
```
CrystalPea, line 71: Let's assert that the report is now marked as completed?

CrystalPea (commented):
Hey Gozie, great work with the test, I left some comments, let me know what you think 🙌🏼

tests/pages/pages.py
BlessedDev: That is a good point. I will implement your suggestion.

tests/pages/pages.py
BlessedDev: Yes. It was a vestige of a different pattern I had wanted to use but forgot to take out.

tests/functional/preview_and_dev/test_unsubscribe_requests.py
BlessedDev: Yes, I couldn't get it to run on preview. I will try again today. If I don't succeed, I think this PR should still be merged and this can be addressed in a separate PR.

CrystalPea (approved):

tests/functional/preview_and_dev/test_unsubscribe_requests.py
BlessedDev: I have added an assertion to check the count of unsubscribe request on the dashboard.

CrystalPea (approved):

risicle:
Only issue is that this doesn't work in dev environments - does it assume a certain pre-existing database state that has been set up by hand on preview/locally?

PR #529, opened 2024-11-01
Use config to get app URLs

quis:
The config setup does some extra work to set the URLs for dev environments:
https://github.com/alphagov/notifications-functional-tests/blob/74282fb66b8b13b3eb978128511f9450c04f2420/config.py#L77-L83

This is missed if the mapping of environment name to URL is used directly.

This causes the test to fail in `dev-a`.

### DIFF
--- a/tests/functional/preview_and_dev/test_unsubscribe_requests.py
+++ b/tests/functional/preview_and_dev/test_unsubscribe_requests.py
@@ -4,7 +4,7 @@ from urllib.parse import urljoin, urlparse
 
 from selenium.webdriver.common.by import By
 
-from config import config, urls
+from config import config
 from tests.pages import (
     DashboardPage,
     UnsubscribeRequestConfirmationPage,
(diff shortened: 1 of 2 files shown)


tests/functional/preview_and_dev/test_unsubscribe_requests.py
```diff
@@ -48,7 +48,7 @@ def test_unsubscribe_request_flow(request, driver, login_seeded_user, client_liv
 
     # simulate an unsubscribe request via the one click unsubscribe url in the body of the email
     path = urlparse(generated_one_click_unsubscribe_url).path
-    admin_url = urljoin(urls[os.environ["ENVIRONMENT"]]["admin"], path)
```
risicle, line 51: I think you need to remove the import of `urls` from the top of the file too.

quis: Thanks, should be fixed now

BlessedDev (approved):

PR #534, opened 2024-11-25
Service join request tests

spatel033:
Add functional tests for service join request story. More details on [this card](https://trello.com/c/izqw07Va/1054-functional-test-for-service-join-request).

### DIFF
--- a/tests/pages/pages.py
+++ b/tests/pages/pages.py
@@ -52,6 +52,7 @@ from tests.pages.locators import (
     VerifyPageLocators,
     ViewLetterTemplatePageLocators,
     ViewTemplatePageLocators,
+    YourServicesPageLocators,
 )
 
 
@@ -326,6 +327,22 @@ class AddServicePage(BasePage):
             element.click()
         except TimeoutException:
             pass
+
+
+class YourServicesPage(BasePage):
+    join_live_service_button = YourServicesPageLocators.JOIN_LIVE_SERVICE_BUTTON
+    add_a_new_service_button = YourServicesPageLocators.ADD_A_NEW_SERVICE_BUTTON
+
+    def wait_until_current(self):
+        return self.wait_until_url_is(self.base_url + "/your-service")
+
+    def join_live_service(self):
+        element = self.wait_for_element(YourServicesPage.join_live_service_button)
+        element.click()
+
+    def add_new_service(self):
+        element = self.wait_for_element(YourServicesPage.add_a_new_service_button)
+        element.click()
 
 
 class ForgotPasswordPage(BasePage):
--- a/tests/pages/pages.py
+++ b/tests/pages/pages.py
@@ -24,6 +24,8 @@ from tests.pages.element import (
     NewPasswordInputElement,
     PasswordInputElement,
     ServiceInputElement,
+    ServiceJoinRequestChooseServiceInputElement,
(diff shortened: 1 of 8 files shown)


risicle:
How does this "reset" itself after the request is accepted the first time?

spatel033:
> How does this "reset" itself after the request is accepted the first time?

It does not "reset" the request, but every time creates new one.

risicle:
But surely on a second or third run the seeded user's already a member of the service and so wouldn't be able to request membership of it. Or is it a different service every time, and if so, doesn't it just continue joining services until it's eventually a member of thousands of services?

spatel033:
> But surely on a second or third run the seeded user's already a member of the service and so wouldn't be able to request membership of it. Or is it a different service every time, and if so, doesn't it just continue joining services until it's eventually a member of thousands of services?

@risicle  For each test run, a new (requester) user account is created. This new user requests access to the "Functional Tests" service, selecting the "Preview admin tests user" as the approver. Subsequently, the test switches to the approver user's account to either approve or reject the access request. The service, organisation, and the specific approver user remain same across all test runs, while the requester user is different every time.

risicle:
Ok, so after running for a couple of months, this service will have hundreds if not thousands of users. That may cause a problem 🤷 it may not. I don't know how (or rather *if*) this is handled in other signup-related tests.

spatel033:
> Ok, so after running for a couple of months, this service will have hundreds if not thousands of users. That may cause a problem 🤷 it may not. I don't know how (or rather _if_) this is handled in other signup-related tests.

Currently we have about 22450 users in preview because we have been registering new users for other tests. We would need separate script to delete redundant data after some time 🤔.

risicle:
Yes maybe we should have such a script.

Or maybe when we get rid of preview and just have staging, we just reset the staging database every week.

tests/pages/pages.py
```diff
@@ -328,6 +335,102 @@ def click_org_type_input(self):
...
 
+class YourServicesPage(BasePage):
+    join_live_service_button = YourServicesPageLocators.JOIN_LIVE_SERVICE_BUTTON
+    add_a_new_service_button = YourServicesPageLocators.ADD_A_NEW_SERVICE_BUTTON
+
+    def wait_until_current(self):
+        return self.wait_until_url_is(self.base_url + "/your-service")
+
+    def join_live_service(self):
+        element = self.wait_for_element(YourServicesPage.join_live_service_button)
```
joybytes, line 346: maybe we should add a call for `self.wait_until_current()` in `join_live_service` and `add_new_service`, otherwise every test would have to invoke the method separately right?

spatel033: `pages` is a common module to define and interact with web elements. [Here we](https://github.com/alphagov/notifications-functional-tests/pull/534/files#diff-6387f8bfb71f1ebd2f1e2165bdf9e7618620fad767aec4be84da05ef5f0a0babR36) `wait_until_current` and then `join_live_service`

joybytes (approved):
I am very happy with the work - well done this was very interesting!

leohemsted:
@spatel033 @risicle 

in the past, we've got a script we've manually run against preview if we started experiencing issues

https://github.com/alphagov/notifications-api/blob/main/app/commands.py#L120

(However, it might need updating for new models, tables, etc - i wouldn't expect us to be able to just run it, but if we needed to it'd provide a launching point)

PR #537, opened 2024-12-02
Install packages using `uv`

quis:
uv is:
- between 10× and 100× faster than pip
- a drop-in replacement for `pip install`, `pip compile`, etc


quis:
Closing this because the lack of virtual environment makes it too difficult to use `uv` both here and locally

PR #540, opened 2024-12-05
always wait before checking current_url

leohemsted:
current_url only reflects what the browser URL is at that point in time

if there's a request in flight, for example you only just clicked a submit button on a form, current_url will still be the old URL. Instead of just asserting on current_url, or making code branch decisions based on it, we should instead use `wait_until_url_contains` which will wait up to ten seconds for the URL to load.

This should hopefully reduce weird flakiness around URL checking and retrieving the service_id from URLs


kr8n3r (approved):
makes sense

PR #541, opened 2024-12-06
Reconfigure test_unsubscribe_request_flow

BlessedDev:
The test made duplicate unsubscribe requests in it's previous form.
 The `one_click_unsubscribe` endpoint now checks for duplicate unsubscribe requests, therefore this PR adjusts the test 
 to now make two separate unsubscribe requests from separate templates with distinct `notification_id`.

### DIFF
--- a/tests/functional/preview_and_dev/test_unsubscribe_requests.py
+++ b/tests/functional/preview_and_dev/test_unsubscribe_requests.py
@@ -16,7 +16,7 @@ from tests.test_utils import create_email_template, go_to_templates_page, record
 
 @recordtime
 def test_unsubscribe_request_flow(request, driver, login_seeded_user, client_live_key):
-    # Create a subscription template
+    # Create subscription template
     go_to_templates_page(driver)
     template_name = f"functional subscription email template {uuid.uuid4()}"
     content = "Hi ((name)), Is ((email address)) your email address? We want to send you some ((things))"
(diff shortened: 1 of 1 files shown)


tests/functional/preview_and_dev/test_unsubscribe_requests.py
```diff
@@ -38,18 +39,37 @@ def test_unsubscribe_request_flow(request, driver, login_seeded_user, client_liv
...
+    resp = client_live_key.post(generated_one_click_unsubscribe_url_1, data={})
     assert resp == {"result": "success", "message": "Unsubscribe successful"}
 
-    dashboard_page.go_to_dashboard_for_service(service_id=config["service"]["id"])
-    dashboard_email_unsubscribe_stats_1 = dashboard_page.get_email_unsubscribe_requests_count()
+    # Create second subscription template
+    go_to_templates_page(driver)
+    template_name_2 = f"functional second subscription email template {uuid.uuid4()}"
+    content = "Hi ((name)), we just wanted to check in!"
+    template_id_2 = create_email_template(driver, name=template_name_2, content=content, has_unsubscribe_link=True)
```
quis, line 53: Do we need a second template? Can we reuse the first one (but still send a new notification)?

quis: It looks like it’s still creating a second template – do you need to push again?

BlessedDev: Good catch. I have pushed the changes now.

quis (approved):

BlessedDev:
@quis I have had to make a change to the placeholder value of the second notification and need a another review.

quis (approved):

PR #543, opened 2024-12-10
Login retry

leohemsted:
### make 2fa sms code logic more robust

i have a theory that it was finding a race condition where the login would fail, but the bit of code where it tries to find the error-message would fail to find the element _before_ the page had reloaded and error message could be rendered.

my gut says that anywhere we say "driver.find_element" without using a `WebDriverWait` could be suceptible to this too, unless we're confident the page has already finished loading.

change the login so that if it cant find the error message, it checks to make sure we're on a different page (which would indicate that we logged in succesfully)

### DIFF
--- a/tests/pages/pages.py
+++ b/tests/pages/pages.py
@@ -188,13 +188,10 @@ class BasePage:
         self.driver.delete_all_cookies()
 
     def wait_until_url_contains(self, url):
-        return WebDriverWait(self.driver, 10).until(self.url_contains(url))
+        return WebDriverWait(self.driver, 10).until(lambda _: url in self.driver.current_url)
 
-    def url_contains(self, url):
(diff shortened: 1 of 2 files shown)


tests/pages/pages.py
```diff
@@ -485,6 +482,22 @@ def verify(self, code):
         self.sms_input = code
         self.click_continue()
 
+    def verify_code_succesful(self) -> bool:
```
tombye, line 485:
```suggestion
    def verify_code_successful(self) -> bool:
```

leohemsted: oops, i've successsfully squashed out that typo

tombye (approved):
I think I've run through all the branches of the new conditions and it makes sense so 👍🏻

PR #544, opened 2024-12-11
fix url checking in verify step

leohemsted:
### always clear events from logging lib

we were only clearing the tests down after failed tests, whereas really it's clearing them down after successful tests that is important

### check for /verify and /two-factor-sms on verify page

verify for new user registration two-factor-sms for returning user sign-in

prev we were just checking for "does url not equal /verify" so got false positives on the user sign-in flow.

also parse url and check path for equality

rather than checking path in a whole URL, split the path.  this means we can be robust (and not accidentally match on URLs that look like `/verify-email` for example), while not getting false negatives if the URL includes, eg, query strings

### DIFF
--- a/tests/pages/pages.py
+++ b/tests/pages/pages.py
@@ -1,7 +1,7 @@
 import os
 import shutil
 from typing import Literal
-from urllib.parse import urlparse
+from urllib.parse import urlparse, urlsplit
 
 from retry import retry
 from selenium.common.exceptions import (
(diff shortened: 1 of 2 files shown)


tests/pages/pages.py
```diff
@@ -498,6 +498,13 @@ def verify_code_successful(self) -> bool:
             # found an error message on the page, so should retry
             return False
 
+    def url_not_a_2fa_sms_page(self):
+        # the page where you enter a 2fa code is one of two URLs:
+        # /verify for user registration, or /two-factor-sms for sign-in
+        WebDriverWait(self.driver, 10).until(
+            lambda _: any(url == self.driver.current_url for url in ("/verify", "/two-factor-sms"))
```
leohemsted: wait this doesnt work hold on

CrystalPea (approved):

PR #545, opened 2024-12-11
uv

leohemsted:
following prior art in eg https://github.com/alphagov/notifications-admin/pull/5311

however, slightly different as our functional tests dockerfile is significantly less complex and not productionised with separate build steps.


Makefile
```diff
@@ -10,13 +10,12 @@ help:
...
 	mkdir -p logs screenshots
-	pip install -r requirements_for_test.txt
+	uv pip install -r requirements_for_test.txt
 
 .PHONY: freeze-requirements
 freeze-requirements: ## create static requirements_for_test.txt
-	pip install --upgrade pip-tools
 	python -c "from notifications_utils.version_tools import copy_config; copy_config()"
-	pip-compile requirements_for_test.in
+	uv compile requirements_for_test.in -o requirements_for_test.txt
```
quis, line 18:
```suggestion
	uv pip compile requirements_for_test.in -o requirements_for_test.txt
```
(at least on my local machine `uv compile` gives `error: unrecognized subcommand 'compile'`

leohemsted: oops yes you're right

quis (approved):

leohemsted:
i need to make sure this works on a dev env with the way they run functional tests first before i merge

PR #547, opened 2024-12-18
Improved logging for multiple failing test cases

spatel033:
Improved logging for multiple failing test cases. Converted module scope to function scope for `test_join_live_service` and `test_registration_and_invite` because driver does not support multiple scopes for logging.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -54,12 +54,14 @@ def _driver(request, download_directory):
     driver.set_window_size(1280, 720)
 
     driver = EventFiringWebDriver(driver, LoggingEventListener())
+    driver._listener.set_node(request.node.name)
 
     driver.delete_all_cookies()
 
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -54,12 +54,14 @@ def _driver(request, download_directory):
     driver.set_window_size(1280, 720)
 
     driver = EventFiringWebDriver(driver, LoggingEventListener())
+    driver._listener.set_node(request.node.name)
```
ipdigital24, line 57: is this `_listener` kind of a private property?

spatel033: `_listner` is the protected property of `EventFiringWebDriver`, and there is no getter method for `_listener`. Currently there is no way to access `set_node`, `print_events` and `clear_events` of our `LoggingEventListener` without using `_listner` and there is no inadvertent side effects, so decided to use the protected property.

PR #549, opened 2024-12-31
Fix functional tests

spatel033:
There are multiple tests are failing when deploying admin. Additionally failing flaky tests cases in preview and also in dev environment. This PRs covers fixing multiple test cases and tested in dev-c environment. More details on [this card](https://trello.com/c/w9hwX6yF/1086-fix-functional-tests).

### DIFF
--- a/tests/functional/preview_and_dev/test_join_live_service.py
+++ b/tests/functional/preview_and_dev/test_join_live_service.py
@@ -1,4 +1,5 @@
 import pytest
+from selenium.common.exceptions import TimeoutException
 
 from config import config
 from tests.pages import (
@@ -6,18 +7,35 @@ from tests.pages import (
     ServiceJoinRequestChoosePermissionsPage,
     ServiceJoinRequestChooseServicePage,
     ServiceJoinRequestJoinAskPage,
+    SignInPage,
     YourServicesPage,
 )
 from tests.pages.rollups import sign_in
 from tests.test_utils import do_user_registration, get_link, recordtime
 
 
-@pytest.fixture(scope="module", autouse=True)
-@recordtime
-def register_user(_driver):
-    # has to use _driver as this is at module level (`driver` fixture is at function level, and just handles taking
-    # the screenshot on failure)
-    do_user_registration(_driver)
+@pytest.fixture(autouse=True)
+def user_register_or_sign_in(driver):
+    sign_in_page = SignInPage(driver)
+    sign_in_page.get()
+
+    try:
+        sign_in_page.wait_until_current()
+    except TimeoutException:
+        # if we didn't get to the sign_in_page, it's probably because we're already logged in.
+        # try logging out before proceeding
+        _do_requester_sign_in(driver)
+    else:
+        do_user_registration(driver)
+
+
+def _do_requester_sign_in(driver):
+    # sign back in as requester (normal user)
+    dashboard_page = YourServicesPage(driver)
+    dashboard_page.sign_out()
+    dashboard_page.wait_until_url_contains(config["notify_admin_url"])
+
+    sign_in(driver, account_type="normal")
 
 
 def _do_approver_sign_in(driver):
@@ -25,10 +43,6 @@ def _do_approver_sign_in(driver):
     sign_in(driver, account_type="seeded")
     invite_link = get_link(config["notify_templates"]["request_invite_to_service_template_id"], requested_user_email)
     driver.get(invite_link)
-
-
-def _do_requester_sign_in(driver):
-    sign_in(driver, account_type="normal")
 
 
 def _do_request_to_join_service(driver):
@@ -65,8 +79,6 @@ def _do_approver_approved_request(driver):
     choose_permissions_page.save_permissions()
     choose_permissions_page.wait_until_url_contains("/users")
 
-    choose_permissions_page.sign_out()
-
 
 def _do_approver_rejected_request(driver):
     _do_approver_sign_in(driver)
@@ -76,8 +88,6 @@ def _do_approver_rejected_request(driver):
     join_service_request_approve_page.continue_to_next_step()
 
     join_service_request_approve_page.wait_until_url_contains("/refused")
-
-    join_service_request_approve_page.sign_out()
 
 
 @recordtime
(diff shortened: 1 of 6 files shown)


joybytes (approved):

CrystalPea (approved):

tests/test_utils.py
```diff
@@ -50,7 +50,7 @@ class NotificationStatuses:
     ACCEPTED = {"accepted"}
     RECEIVED = {"received"}
     DELIVERED = {"delivered", "temporary-failure", "permanent-failure"}
-    SENT = RECEIVED | DELIVERED | {"sending", "pending"}
+    SENT = RECEIVED | DELIVERED | {"sending", "pending", "created"}
```
leohemsted, line 53: why is the status still in `created`? is it stuck in our system? if so, we need to wait for it to progress

can you give an example of a notification ID on preview that was stuck in created? how long was it stuck for? was it from a job, or from api? etc?

i don't buy "sometimes things for mmg are stuck in created" - as soon as we send to MMG, we must update immediately to `sending` or notify is severely broken

tests/pages/pages.py
```diff
@@ -762,6 +762,7 @@ def click_send_button(self):
 
     def click_change_language(self):
         element = self.wait_for_element(self.change_language_button)
+        self.driver.execute_script("window.scrollTo(0, 50)")
```
leohemsted, line 765: cc @tombye i think this is the one here:
image

dont know if we have a bug for this in trello?

tests/functional/preview_and_dev/test_unsubscribe_requests.py
```diff
@@ -85,6 +86,7 @@ def test_unsubscribe_request_flow(request, driver, login_seeded_user, client_liv
 
     # Download email unsubscribe request report
     report_page = UnsubscribeRequestReportPage(driver)
+    time.sleep(10)
```
leohemsted, line 89: i don't love random sleeps - what's happening here?

Do you have a link to a concourse job that failed at this point or a screenshot or something? cant figure out how `report_page.click_download_report_link()` would result in a different result without this sleep.

the click function should wait up to ten seconds anyway - if these unsubscribe reports take more than twenty seconds to load, that's an issue in of itself

tests/functional/preview_and_dev/test_registration_and_invite.py
```diff
@@ -42,6 +33,8 @@ def _sign_in_again(driver):
 @recordtime
 @pytest.mark.xdist_group(name="registration-flow")
 def test_invite_flow(driver):
+    do_user_registration(driver)
+    do_user_add_new_service(driver)
```
leohemsted, line 37: this means that we won't have user registration or create the service if we just run `test_can_add_and_update_reply_to` or `test_can_add_and_update_sms_sender_of_service` 🤔 

is that ok? is this necessary?

leohemsted (changes requested):

tests/test_utils.py
klssmith: I think MMG messages are staying in `created` because we're getting an error when sending to them on dev environments due to having run out of credit. It looks like we only update to `sending` if there were no exceptions raised when sending. We are just looking at increasing the credit though, which should mean we don't need to add "created" here (see https://gds.slack.com/archives/C03QVEHD4EA/p1734616273586009 for details)

leohemsted: ahhh thanks for the context. yes i'd rather we fix the issues with credit - we rely on these enums for everything incl smoke tests. if we can't send by MMG, that's a test failure.

tests/pages/pages.py
spatel033: I think it was part of [this card](https://trello.com/c/wPRphm25/998-the-sticky-footer-sometimes-overlaps-focused-elements?search_id=5cadd11e-418e-49f2-9f71-008499f6c48a). Dropped the changes for now.

tests/functional/preview_and_dev/test_registration_and_invite.py
spatel033: Good point!! Refactored the fixture for the function scope

tests/functional/preview_and_dev/test_unsubscribe_requests.py
spatel033: Few times replicated in dev environment, will need to investigate further. For now removing this change.

tests/test_utils.py
spatel033: Thanks @klssmith. Since this is Dev only issue, will create ticket for this to fix the issue for credit related status update.

tests/functional/preview_and_dev/test_join_live_service.py
```diff
@@ -1,23 +1,40 @@
...
+    sign_in_page = SignInPage(driver)
+    sign_in_page.get()
+
+    try:
+        sign_in_page.wait_until_current()
+        do_user_registration(driver)
+    except TimeoutException:
+        # if we didn't get to the sign_in_page, it's probably because we're already logged in.
+        # try logging out before proceeding
+        _do_requester_sign_in(driver)
```
leohemsted, line 28: im wondering what happens if a timeout happens during `do_user_registration` - i dont think we should be trying to sign in in that case - just checking the sign_in_page should be enough I think? So I think this is a bit safer:

```python
    try:
        sign_in_page.wait_until_current()
    except TimeoutException:
        # if we didn't get to the sign_in_page, it's probably because we're already logged in.
        # try logging out before proceeding
        _do_requester_sign_in(driver)
    else:
        do_user_registration(driver)
```

tests/functional/preview_and_dev/test_registration_and_invite.py
```diff
@@ -18,13 +20,19 @@
...
+    sign_in_page.get()
+
+    try:
+        sign_in_page.wait_until_current()
+        do_user_registration(driver)
+        do_user_add_new_service(driver)
+    except TimeoutException:
+        # if we didn't get to the sign_in_page, it's probably because we're already logged in.
+        # try logging out before proceeding
+        _sign_in_again(driver)
```
leohemsted, line 35: same comment as above - i think we can use `else` here to only run registration/service creation if the user isn't signed up and it doesnt raise an exception

```python
    try:
        sign_in_page.wait_until_current()
    except TimeoutException:
        # if we didn't get to the sign_in_page, it's probably because we're already logged in.
        # try logging out before proceeding
        _sign_in_again(driver)
    else:
        do_user_registration(driver)
        do_user_add_new_service(driver)
```

leohemsted (approved):

PR #553, opened 2025-01-09
wait for unsubscribe email to move to sending

leohemsted:
use the same pattern we do elsewhere and wait for the notification to move into a sending/delivered status before clicking the unsubscribe link

we saw a race condition where we were downloading an unsubscribe report that contained an email that hadn't been sent yet (possibly because the celery apps on dev envs are slow to start up). We rely on the notification.sent_at field for one of the columns in the report. This is set when the email is sent by celery, but because we're getting the URL via the public api we can get there too early.


quis (approved):
Good to go once Concourse is happy

PR #560, opened 2025-02-20
Fix inbound sms

spatel033:


leohemsted (commented):
looks good but please update the commit messages and PR description to clearly explain what you're doing and why it's necessary

leohemsted (approved):
nice one. checked the numbers in the database and they're all mixed. think this'll work alright

PR #566, opened 2025-04-09
upgrade setuptools

leohemsted:
the official base python images now ship with incompatible versions of wheel and setuptools, which can cause issues with running `python setup.py bdist_wheel`, which happens for some of our dependencies and subdependencies.

For now, a workaround is to bump setuptools to a version that doesn't have this incompatibility. We should remove this line in future once the base python image has been updated

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,6 +19,9 @@ WORKDIR /var/project
 
 COPY . .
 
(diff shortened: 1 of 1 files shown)


quis (approved):
After removing my `python:3.11` image locally I can confirm:
- `main` is broken in the same way we are seeing on Concourse 
- this branch fixes the problem

PR #569, opened 2025-04-22
Update button text for join a service

klssmith:
We're changing the text on the button for joining a service from "Join a live service" to "Join an existing service". This change allows the tests to pass with either text while the change is deployed. After, we can remove references to the old button text.

### DIFF
--- a/tests/pages/pages.py
+++ b/tests/pages/pages.py
@@ -339,6 +339,7 @@ class AddServicePage(BasePage):
 
 
 class YourServicesPage(BasePage):
+    join_existing_service_button = YourServicesPageLocators.JOIN_EXISTING_SERVICE_BUTTON
     join_live_service_button = YourServicesPageLocators.JOIN_LIVE_SERVICE_BUTTON
     add_a_new_service_button = YourServicesPageLocators.ADD_A_NEW_SERVICE_BUTTON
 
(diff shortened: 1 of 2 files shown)


kr8n3r (approved):

tests/pages/pages.py
```diff
@@ -339,14 +339,18 @@ def click_org_type_input(self):
...
+    join_existing_service_button = YourServicesPageLocators.JOIN_EXISTING_SERVICE_BUTTON
     join_live_service_button = YourServicesPageLocators.JOIN_LIVE_SERVICE_BUTTON
     add_a_new_service_button = YourServicesPageLocators.ADD_A_NEW_SERVICE_BUTTON
 
     def wait_until_current(self):
         return self.wait_until_url_contains(self.base_url + "/your-services")
 
     def join_live_service(self):
-        element = self.wait_for_element(YourServicesPage.join_live_service_button)
+        try:
```
leohemsted, line 350: can we add a `# TODO` comment so we remember to remove this once we've made the code change

klssmith: I'm already writing the code to remove it, so will probably be able to get rid of this extra code today

klssmith: @leohemsted this is the PR to remove it https://github.com/alphagov/notifications-functional-tests/pull/570

PR #573, opened 2025-04-30
Increase retries on assert_callback_received

corlettb:
https://trello.com/c/koHryaZa/1092-make-the-functional-tests-work-in-dev-environments

What
----

- Increase default retries and delay on assert_callback_received
- Fix config 'dev-*' overrides for assert_callback_received. The function decorators are being applied before the `@pytest.fixture(scope="session", autouse=True)` executes and so we aren't picking up the overrides

Why
----

I have seen a couple of failures here over the last couple of days. See:

https://concourse.notify.tools/teams/dev-a/pipelines/deploy-notify/jobs/functional-tests/builds/135#L67f96f92:105
https://concourse.notify.tools/teams/dev-a/pipelines/deploy-notify/jobs/functional-tests/builds/120#L67f936ae:105

I investigated the first and found the event did make it to pipedream:

```
    {
      "id": "1746009508480-0",
      "indexed_at_ms": 1746009508480,
      "event": {
        ...
        "body": {
          "id": "8254df49-dd3f-4cdc-8abd-85b7e34ef969",
          "source_number": "447700900345",
          "destination_number": "07700900500",
          "message": "Inbound message from dev-a_Functional Test_9ee281f2-0b6c-4eaf-bb00-e21183d1a3e0",
          "date_received": "2025-04-30T10:38:20.000000Z"
        }
      },
      ...
    },
```

Comparing the indexed_at_ms with the date_received:

```
    date_received: 2025-04-30T10:38:20.000000Z
    indexed at: Wednesday, 30 April 2025 10:38:28.480
```

It looks like it took pipedream 8 seconds to index it which could well explain why the test failed.


risicle:
Is the "fix" needed because a properly populated `config` isn't available at import-time?

corlettb:
> Is the "fix" needed because a properly populated `config` isn't available at import-time?

Yeah ordering issue. You get the default config. So we only get 5 retries with a delay of 0.5 seconds. Not enough.

risicle (approved):

PR #643, opened 2026-03-26
Add test sending a file one off via document download via UI

BlessedDev:
This PR adds a test covering sending a file one off via document download via UI.
It is dependant on this [PR](https://github.com/alphagov/notifications-functional-tests/pull/636)

### DIFF
--- a/tests/notifications/functional_tests/test_document_download_via_ui.py
+++ b/tests/notifications/functional_tests/test_document_download_via_ui.py
@@ -1,6 +1,20 @@
 import uuid
 
-from tests.pages import ManageFilesForEmailTemplatePage, ShowTemplatesPage, ViewEmailTemplatePage
+import pytest
+
+from config import config
+from tests.pages import (
+    ChangeLinkTextForEmailFilePage,
+    DocumentDownloadLandingPage,
+    ManageEmailTemplateFilePage,
+    ManageFilesForEmailTemplatePage,
+    SendEmailPreviewPage,
+    SendOneRecipientPage,
+    SendSetSenderPage,
+    SentEmailMessagePage,
+    ShowTemplatesPage,
+    ViewEmailTemplatePage,
+)
 from tests.test_utils import (
     create_an_email_template_and_attach_a_file,
     delete_file_from_email_template_via_manage_files_page,
@@ -9,6 +23,7 @@ from tests.test_utils import (
 
 
 @recordtime
+@pytest.mark.xdist_group(name="send-files-via-ui-flow")
 def test_attaching_files_to_emails_and_also_deleting_them_via_ui(driver, login_seeded_user):
     # Test Creating an email template and attach a file to it
     template_name = f"Functional Tests - upload/delete file to email template - {uuid.uuid4()}"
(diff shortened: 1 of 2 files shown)


tests/notifications/functional_tests/test_document_download_via_ui.py
```diff
@@ -1,6 +1,9 @@
 import uuid
 
-from tests.pages import ViewEmailTemplatePage, ManageFilesForEmailTemplatePage, ShowTemplatesPage
+from config import config
```
CrystalPea, line 5: Related to a discussion I started here: https://github.com/alphagov/notifications-api/pull/4784/changes#r3027240483

This file is called `test_document_download_via_ui`, but this is not the words we would use to describe this journey to our users. So, while possibly not in this PR, as it is interconnected with other functional tests PR, I would be for renaming this file `test_send_files_by_email_via_ui` - which describes the action user would do better.

tests/notifications/functional_tests/test_document_download_via_ui.py
```diff
@@ -41,3 +44,85 @@ def test_attaching_files_to_emails_and_also_deleting_them_via_ui(driver, login_s
     templates_page = ShowTemplatesPage(driver)
     assert templates_page.get_h1_text() == "Templates"
     assert template_name not in templates_page.get_all_listed_templates()
+
+
+@recordtime
+def test_sending_a_file_one_off_via_document_download_via_ui(driver, login_seeded_user):
```
CrystalPea: [suggestion] We could use user-facing terminology here and make the function name more succinct at the same time: `test_send_one_off_email_with_file_via_ui`

tests/notifications/functional_tests/test_document_download_via_ui.py
```diff
@@ -41,3 +44,85 @@ def test_attaching_files_to_emails_and_also_deleting_them_via_ui(driver, login_s
...
+    # Create an email template
+    template_name = f"Functional Tests - send a file off one off via ui - {uuid.uuid4()}"
```
CrystalPea:
```suggestion
    template_name = f"Functional Tests - send one off email with file via ui - {uuid.uuid4()}"
```

tests/notifications/functional_tests/test_document_download_via_ui.py
```diff
@@ -41,3 +44,85 @@ def test_attaching_files_to_emails_and_also_deleting_them_via_ui(driver, login_s
...
+    content = "Hi download this file:"
+    file_name = "attachment.pdf"
```
CrystalPea, line 67: I looked through other template contents, and they usually say what they are testing - maybe we could repeat that here?

For example:

```
    content = "Testing sending a one off email notification. with an email file. Test file below:"
    file_name = "functional_tests_email_file.pdf"
```

tests/pages/pages.py
```diff
@@ -208,6 +211,10 @@ def sign_out(self):
 
         self.driver.delete_all_cookies()
 
+    def click_templates_link(self):
```
CrystalPea: an equivalent function already exists for this class ('click_templates()`) so this is a duplicate.

tests/pages/pages.py
```diff
@@ -943,6 +957,26 @@ def click_remove_file_dialog_button(self):
...
+        )
+        element = self.wait_for_element((By.XPATH, xpath))
+        element.click()
+
+
+class ChangeLinkTextForEmailFilePage(BasePage):
+    link_text_input = (By.CSS_SELECTOR, "input[name='link_text'][id='link_text']")
+    continue_button = (By.CSS_SELECTOR, "button[type='submit']")
+
+    def click_continue_button(self):
```
CrystalPea, line 968: `BasePage` class, which this one inherits from, already has `click_continue()` function, so this one is a duplicate.

tests/pages/pages.py
```diff
@@ -1470,6 +1505,10 @@ def choose_alternative_sms_sender(self):
         radio = self.wait_for_design_system_checkbox_or_radio(self.alternative_sender_sms_radio)
         radio.click()
 
+    def click_continue_button(self):
```
CrystalPea, line 1503: also a duplicate.

tests/pages/pages.py
```diff
@@ -1511,6 +1550,26 @@ def click_use_emergency_list(self):
         element.click()
 
 
+class SendEmailPreviewPage(BasePage):
```
CrystalPea, line 1548: do we need this class? We have another test where we are sending one off email, and we haven't needed it there: https://github.com/alphagov/notifications-functional-tests/blob/main/tests/test_utils.py#L408

tests/pages/pages.py
```diff
@@ -1511,6 +1550,26 @@ def click_use_emergency_list(self):
...
+    send_button = (By.CSS_SELECTOR, "button[type='submit'")
+
+    def click_send_button(self):
+        element = self.wait_for_element(SendEmailPreviewPage.send_button)
+        element.click()
+
+
+class SendEmailConfirmationPage(BasePage):
```
CrystalPea: class name makes it sound like we are confirming before sending, but I understand this is the sent notification page maybe we could rename it to `EmailNotificationPage`, as the object we are dealing with at this point is a notification? Or `SentEmailMessagePage` if we want to use user-facing wording?

tests/pages/pages.py
```diff
@@ -1795,3 +1854,24 @@ def select_mark_as_complete_checkbox(self):
     def click_update_button(self):
         element = self.wait_for_element(UnsubscribeRequestReportPage.update_report_button)
         element.click()
+
+
+class YouHaveAFileToDownloadPage(BasePage):
```
CrystalPea: I think this and next class might be duplicates of classes for these existing tests? 

https://github.com/alphagov/notifications-functional-tests/blob/c99aeb39212d191579da84ad68257ccb2639948d/tests/document_download/functional_tests/test_document_download.py

I like your names better though :) - maybe we could rename existing classes and remove duplicates?

CrystalPea (commented):
Good (and extensive) work, I left some comments - mainly renaming discussion/suggestions, and spotted some duplicates 🙌🏼

tests/notifications/functional_tests/test_document_download_via_ui.py
BlessedDev: Point taken. I will look at renaming the file in another PR.

tests/pages/pages.py
BlessedDev: I had to override the method in the Base class which uses a different selector which didn't work in this case.  There are slight variations across pages as per the continue button that means that the Base class method doesn't work in all circumstances

tests/pages/pages.py
BlessedDev: Yes. I tried using the the method you refereed to  and it didn't work.
There were slight nuances in the journey, as compared to those that the method was originally designed for. It is one of the reasons I have added assertions as the test transitions from page to page to indicate the steps in the journey.

tests/pages/pages.py
BlessedDev: `SentEmailMessagePage` sounds good to me.

tests/pages/pages.py
BlessedDev: Good point. I will remove the new duplicate classes.
I can see how the original classes would work better especially to give a greater contrast with the classes I am working on for the Preview mirrors of the document download pages in admin.

tests/notifications/functional_tests/test_document_download_via_ui.py
BlessedDev: I have updated the content. However I have retained the original file I had used because it is part of the FT test corpus https://github.com/alphagov/notifications-functional-tests/blob/main/tests/test_files/attachment.pdf

tests/pages/pages.py
CrystalPea: I wonder if your new selector would work for the existing tests - then we could still use one method for both?

tests/pages/pages.py
CrystalPea: That's fair - in that case, it would be worth adding that context - why we have a separate class - maybe as a docstring for the class? For example

"""
This journey for sending one-off email is slightly different from our main one-off email test. Because of this, the existing page class didn't work for it, and this new class had to be created.
"""

tests/notifications/functional_tests/test_document_download_via_ui.py
```diff
@@ -41,3 +44,85 @@ def test_attaching_files_to_emails_and_also_deleting_them_via_ui(driver, login_s
...
+    assert preview_send_one_recepient_page.get_h1_text() == f"Preview of ‘{template_name}’"
+    preview_send_one_recepient_page.click_send_button()
+
+    # confirm that the email is being delivered
+    send_email_confirmation_page = SentEmailMessagePage(driver)
+    assert send_email_confirmation_page.get_h1_text() == "Email"
+    assert "Delivering" in send_email_confirmation_page.get_notification_status()
+
+    # Confirm that the file download link sent to the recipient works
+    # There are smoke tests and other tests covering the process of a recipient downloading
```
CrystalPea, line 118: are there smoke tests for this? I thought we only had functional tests for it?

BlessedDev: There are smoke tests as well https://github.com/alphagov/notifications-functional-tests/blob/main/tests/document_download/smoke_tests/test_document_download.py

CrystalPea: Oh wow, today I learnt, thanks!

tests/pages/pages.py
BlessedDev: Yes, I will add comments to the class.

tests/pages/pages.py
BlessedDev: Not all buttons are of the `type` , `submit`. There are differences across various pages depending on wether we are dealing with forms, actual button elements or anchor elements with the type as button.

I think we can maintain the base class method and override it within individual page classes if needed.

CrystalPea: Ahh, that is interesting. There would be a few categories all buttons would fit into though, right? So if we have repetitions of the same category, it would be better to create an additional BaseClass method for this and reuse it /make existing one more versatile rather than having lots of repeated code - DRY principle is important for ease of maintanance.

If this is the first instance where our continue button is different though, that's ok.

CrystalPea (approved):

CrystalPea (approved):

quis (approved):

PR #653, opened 2026-04-20
Add test_send_email_notification_with_an_email_file_via_csv

BlessedDev:
We had to revert the previously deployed Functional Tests for sending email notifications with files attached via `csv`, because they were unexpectedly flaky.

In  this PR, the test does not use the existing `send_notification_via_csv` used by other FTs. The plan is to refactor the code in subsequent PRs and address the slight nuances that meant the existing code used in by other tests hasn't worked in the send file vi ui tests.

Also subsequent PRs will refactor all the send file via ui FTs to keep things DRY. The priority is to get the test working so that the feature can be deployed.

I have added comments to try to describe the reasoning for not using some of the existing code and patterns used in other FTs (ie not send file via ui)

### DIFF
--- a/tests/notifications/functional_tests/test_document_download_via_ui.py
+++ b/tests/notifications/functional_tests/test_document_download_via_ui.py
@@ -1,9 +1,8 @@
 import uuid
 
 import pytest
-
-from config import config
 from pypdf import PdfReader
+from selenium.webdriver.common.by import By
 
 from config import config, generate_unique_email
 from tests.pages import (
@@ -11,24 +10,28 @@ from tests.pages import (
     ChangeRentionPeriodForEmailFilePage,
     DocumentDownloadLandingPage,
     EmailConfirmationSettingForEmailFilePage,
+    JobPage,
     ManageEmailTemplateFilePage,
     ManageFilesForEmailTemplatePage,
-    SendEmailPreviewPage,
-    SendOneRecipientPage,
-    SendSetSenderPage,
-    SentEmailMessagePage,
     PreviewConfirmYourEmailAddressPage,
     PreviewDownloadYourFilePage,
     PreviewYouHaveAFileToDownloadPage,
+    SendEmailPreviewPage,
+    SendFilesViaUiUploadCsvPage,
+    SendOneRecipientPage,
+    SendSetSenderPage,
+    SendViaCsvPage,
+    SendViaCsvPreviewPage,
+    SentEmailMessagePage,
     ShowTemplatesPage,
-    ViewEmailTemplatePage, SendViaCsvPage, SendViaCsvPreviewPage, JobPage,
+    ViewEmailTemplatePage,
 )
-from tests.postman import send_notification_via_csv
 from tests.test_utils import (
     create_an_email_template_and_attach_a_file,
+    create_temp_csv,
     delete_file_from_email_template_via_manage_files_page,
     get_downloaded_document,
-    recordtime, create_temp_csv,
+    recordtime,
 )
 
 
@@ -137,6 +140,7 @@ def test_send_one_off_email_with_file_via_ui(driver, login_seeded_user):
     # delete the template
     assert view_email_template_page.get_h1_text() == template_name
     delete_email_template_for_send_file_via_ui_tests(driver, view_email_template_page, template_name)
+
 
 @recordtime
 @pytest.mark.xdist_group(name="send-files-via-ui-flow")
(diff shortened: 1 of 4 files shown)


tests/notifications/functional_tests/test_document_download_via_ui.py
```diff
@@ -315,3 +316,93 @@ def test_send_file_via_ui_preview_pages(driver, login_seeded_user, download_dire
...
+    status = send_email_confirmation_page.get_notification_status()
+    assert "Delivering" in status or "Delivered"  # Either status pops up, depending on the test run
+
+    # Confirm that the file download link sent to the recipient works
+    # There are smoke tests and other tests covering the process of a recipient downloading
+    # a document, so the whole journey will not be covered here
+    send_email_confirmation_page.click_file_download_link(link_text)
+    you_have_a_file_to_download_page = DocumentDownloadLandingPage(driver)
+    assert you_have_a_file_to_download_page.get_h1_text() == "You have a file to download"
+    you_have_a_file_to_download_page.go_to_download_page()
```
rparke: Small one - can we confirm the URL goes to document download? Or goes to the page without the banner. If it's not a trivial add, we can add it later.

BlessedDev: It is not a difficult check to add. I will add it.

BlessedDev: To clarify, I will quickly add a check that it gets to the actual download page but will not go through with the actual download which is already covered by the smoke tests and document download tests.

tests/notifications/functional_tests/test_document_download_via_ui.py
```diff
@@ -59,13 +59,7 @@ def test_attaching_files_to_emails_and_also_deleting_them_via_ui(driver, login_s
...
     # delete template
     assert view_email_template_page.get_h1_text() == template_name
-    view_email_template_page.click_delete_template_link()
-    view_email_template_page.click_template_deletion_confirmation_button()
-
-    # confirm template has been deleted
-    templates_page = ShowTemplatesPage(driver)
-    assert templates_page.get_h1_text() == "Templates"
-    assert template_name not in templates_page.get_all_listed_templates()
+    delete_email_template_for_send_file_via_ui_tests(driver, view_email_template_page, template_name)
```
rparke, line 65: ultimately I wonder if we should try and add these to a `try` `finally` block or something. It's a small thing but it stops us from accumulating "stale" templates if the tests flake.

Not a priority for this though

BlessedDev: I am considering something similar for an upcoming refactoring of all the send files via ui FTs.
It can be addressed in subsequent PRs

tests/notifications/functional_tests/test_document_download_via_ui.py
BlessedDev: Thinking about this further. I will have a quick go at going all the way. It does seem incomplete without the download even though there are the aforementioned smoke tests etc.

rparke: I don't think we need to download it. It was just a check that we successfully added a link that goes to document download vs the preview page.

It's more places for the test to be flakey.

tests/notifications/functional_tests/test_document_download_via_ui.py
rparke: cool, can consider it resolved it was just a point I wanted to make for later.

tests/notifications/functional_tests/test_document_download_via_ui.py
```diff
@@ -308,10 +294,103 @@ def test_send_file_via_ui_preview_pages(driver, login_seeded_user, download_dire
...
+    assert set_sender_page.get_h1_text() == "Where should replies come back to?"
+    set_sender_page.click_continue_button()
+
+    send_via_csv_page = SendViaCsvPage(driver)
+    assert send_via_csv_page.get_h1_text() == f"Send ‘{template_name}’"
+    # create temp csv for this test
+    send_via_csv_page.go_to_upload_csv_for_service_and_template(config["service"]["id"], template_id)
+    upload_csv_page = SendFilesViaUiUploadCsvPage(driver)
+    assert send_via_csv_page.get_h1_text() == "Upload a list of email addresses"
+    seeded_user_email = generate_unique_email(
```
BlessedDev, line 349: There has been an issue where `config["service"]["seeded_user"]["email"]` is not acceptable. Hence the need to generate a csv file for this test.

tests/notifications/functional_tests/test_document_download_via_ui.py
```diff
@@ -308,10 +294,103 @@ def test_send_file_via_ui_preview_pages(driver, login_seeded_user, download_dire
...
+        config["service"]["seeded_user"]["email"], "test_send_file_via_ui_preview_pages"
+    )
+    _, directory, csv_filename = create_temp_csv({"email address": seeded_user_email}, include_build_id=True)
+    upload_csv_page.upload_csv(directory, csv_filename)
```
BlessedDev, line 353: Also deviated from the established `upload_csv` patterns for other tests as it didn't completely work with the `send file via csv flow`.

Given the time constraints, it was easier to have a solution for these tests and then to look at reconciling the differences in subsequent PRs.

rparke (approved):

PR #654, opened 2026-04-23
Refactor send files via UI tests

BlessedDev:
This PR does some refactoring and clean up of the `send file via ui ` FTs.
I have merged the sending "one-off" and "via csv" tests.


tests/test_utils.py
```diff
@@ -725,3 +725,12 @@ def change_email_template_email_confirmation(driver, email_confirmation_label, m
     assert email_confirmation_page.get_h1_text() == "Ask recipient for their email address"
     email_confirmation_page.select_email_confirmation_option(new_confirmation_label_choice)
     email_confirmation_page.click_continue_button()
+
+
+def go_back_to_email_template_from_file_management_page(file_name, manage_a_file_page, template_name,
```
CrystalPea: [suggestion] we could change the function name like the other helper - so it matches how we call the involved pages elsewhere in the tests.

tests/notifications/functional_tests/test_send_files_via_ui.py
```diff
@@ -65,63 +51,67 @@ def test_attaching_a_file_to_an_email_template_and_also_removing_the_file(driver
...
+    elif recipient_type == "csv_of_recipients":
+        send_email_notification_with_file_attached_via_csv(driver, template_id, template_name, recipient_data)
 
     # Confirm that the email is being delivered
     send_email_confirmation_page = SentEmailMessagePage(driver)
-    assert_email_notification_with_a_file_attached_was_sent(send_email_confirmation_page)
+    assert send_email_confirmation_page.get_h1_text() == "Email"
+    status = send_email_confirmation_page.get_notification_status()
+    assert (
+        "Delivering" in status or "Delivered"
```
CrystalPea: might be clearer to write it as: `assert status in ["Delivering", "Delivered"]`

tests/notifications/functional_tests/test_send_files_via_ui.py
```diff
@@ -0,0 +1,304 @@
...
+    go_to_email_template_preview_page,
+    recordtime,
+    send_email_notification_with_file_attached_to_one_recipient,
+    send_email_notification_with_file_attached_via_csv,
+)
+
+
+@recordtime
+@pytest.mark.xdist_group(name="send-files-via-ui-flow")
+def test_attaching_a_file_to_an_email_template_and_also_removing_the_file(driver, login_seeded_user):
```
CrystalPea: [nitpick] we usually do the test names in present simple, for example test_attach_a_file_to_an_email_template_and_then_remove_the_file

do we want to change it to match? Same on line 66 🙌🏼

tests/notifications/functional_tests/test_send_files_via_ui.py
```diff
@@ -0,0 +1,304 @@
...
+    # Creat an email template and attach a file to it
```
CrystalPea: [typo] this would need fixing throughout the file

```suggestion
    # Create an email template and attach a file to it
```

tests/test_utils.py
```diff
@@ -321,6 +326,13 @@ def go_to_templates_page(driver, service="service"):
     dashboard_page.click_templates()
 
 
+def go_to_email_template_preview_page(driver, template_name):
```
CrystalPea: template preview is a bit of a confusing name, as it's also a name of one of our apps. Usually we call that page `View Template Page`, so maybe we could rename this template for readability?

Also the helper will work for any type of template, so maybe we could name it like this:

```suggestion
def go_to_view_template_page(driver, template_name):
```

tests/test_utils.py
```diff
@@ -647,42 +660,97 @@ def pdf_page_has_text(pdf_page, expected_text, normalise_whitespace=True):
     return expected_text in page_text
 
 
+def delete_file_from_email_template_via_manage_files_page(driver):
+    manage_a_file_page = ManageEmailTemplateFilePage(driver)
+    manage_a_file_page.click_remove_file_link()
+    manage_a_file_page.click_remove_file_dialog_button()
+
+
+def go_to_email_template_preview_page_from_a_document_download_pages(driver, template_name, document_download_page):
```
CrystalPea: Are we using this anywhere?  I can't find it being called in this PR.

tests/notifications/functional_tests/test_send_files_via_ui.py
```diff
@@ -0,0 +1,304 @@
...
+    expected_text = " ".join("This is an attachment").split()
+    assert result_text == expected_text
+
+    # go to email template preview page and delete the template
+    go_to_email_template_preview_page(driver, template_name)
+    delete_email_template_from_email_template_preview_page(driver, template_name)
+
+
+# Send file via ui tests helper methods to keep things DRY
+def go_to_file_management_page_from_email_template_preview(driver, file_name):
```
CrystalPea: [suggestion] maybe we could slightly rename that function so it matches how we call these pages in all the variables - makes reading a bit easier:

```suggestion
def go_to_manage_file_page_from_view_email_template_page(driver, file_name, view_email_template_page):
```

tests/notifications/functional_tests/test_send_files_via_ui.py
```diff
@@ -0,0 +1,304 @@
...
+    assert change_link_text_page.get_h1_text() == "Add link text"
+    change_link_text_page.fill_in_link_text(new_link_text)
+    change_link_text_page.click_continue_button()
+    # confirm link text change
+    assert manage_a_file_page.get_h1_text() == file_name
+    assert manage_a_file_page.get_file_setting_value(link_text_label) == new_link_text
+    return change_link_text_page
+
+
+def change_email_template_retention_period(
```
CrystalPea: [question] hmm I see how changing link text benefits from being extracted into a helper function as we do it multiple times.

But we only change retention period and email confirmation once - what is the reason we are extracting those?

CrystalPea (commented):
Hi there, good work on the refactor!

I have some suggestions:

- some helpers are in the new `test_send_files_via_ui.py‎`, while some are in utils file, even though they are only used by the new file - should we move all the helpers to `test_send_files_via_ui.py‎`, unless they are shared across multiple files?
-  some helpers are only used once - are all of those worth extracting into separate functions? I think worth considering where it will help with readability, and where it will hinder it.
- when naming the helpers, as much as possible, lets keep names for pages consistent with how we call them in the variables - that will help the person reading the test understand what is happening at a glance.

Those are general notes - I also left a few comments to illustrate these, and highlight a couple other small issues 🙌🏼

tests/notifications/functional_tests/test_send_files_via_ui.py
BlessedDev: No problem, I will rename the test.

tests/notifications/functional_tests/test_send_files_via_ui.py
BlessedDev: Good catch, I will correct the typo.

tests/test_utils.py
BlessedDev: I take your point, I will rename it as suggested.

tests/test_utils.py
BlessedDev: No. I had replaced it with `go_to_view_template_page` and forgot to take it out. I will remove it.

tests/notifications/functional_tests/test_send_files_via_ui.py
BlessedDev: I had initially done that as I wanted to add more test cases but thought against it in the end but left these in incase of future tests.
However I don't feel strongly about leaving them in, so I will remove them.

BlessedDev:
> Hi there, good work on the refactor!
> 
> I have some suggestions:
> 
> * some helpers are in the new `test_send_files_via_ui.py‎`, while some are in utils file, even though they are only used by the new file - should we move all the helpers to `test_send_files_via_ui.py‎`, unless they are shared across multiple files?
> * some helpers are only used once - are all of those worth extracting into separate functions? I think worth considering where it will help with readability, and where it will hinder it.
> * when naming the helpers, as much as possible, lets keep names for pages consistent with how we call them in the variables - that will help the person reading the test understand what is happening at a glance.
> 
> Those are general notes - I also left a few comments to illustrate these, and highlight a couple other small issues 🙌🏼

This was a tricky one. I had split out the functions that I thought could be used independently of these tests into the `test_utils.py` file for future tests. 
However perhaps I was looking to far ahead. I will move them back into `test_send_files_via_ui.py`.

CrystalPea (approved):

PR #656, opened 2026-05-18
`client_live_key`, `client_test_key` fixtures: provide `client_id` for correlation

risicle:
Functioning similarly to `driver_id` and included in `User-agent`, this should make it possible to correlate requests from a particular FT process with its corresponding logs server-side.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,7 +3,6 @@ from datetime import datetime
 from pathlib import Path
 
 import pytest
-from notifications_python_client import NotificationsAPIClient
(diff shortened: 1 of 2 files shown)


risicle:
A slight complication because `get_notification_by_to_field` uses its own api client instance which doesn't immediately seem to fit well into a fixture. Instead it's going to get its own client (and therefore client id) per call. But in many of these cases the stack trace helpfully contains the headers dict that shows the `User-agent`, meaning not having display-on-teardown isn't a complete loss.

PR #660, opened 2026-06-09
Change test name, correct two comments, and add two TODOs

CrystalPea:


tests/pages/pages.py
```diff
@@ -592,10 +592,12 @@ def get_service_name(self):
         element = self.wait_for_element(DashboardPage.h2)
         return element.text
 
+    # TODO: not used, delete?
```
joybytes: non-blocking: maybe we can delete these now if they are not in use?

joybytes (approved):

tests/pages/pages.py
CrystalPea: good idea, I will do that!

joybytes (approved):

PR #663, opened 2026-07-17
fix: flakey test test_template_folder_permissions

joybytes:
### Summary

fix test_template_folder_permissions which I saw failing here: https://concourse.notify.tools/teams/staging/pipelines/deploy-notify/jobs/functional-tests/builds/1952

Team members: /services/S/users
- Edit team member: /services/S/users/U (uncheck a folder permission checkbox)
- Click Save → form POST → redirects back to /services/S/users

Without the wait, the test clicked the "Team members" nav link immediately after clicking Save. 
It went away while the save POST was still running made the browser cancel the POST, so the permission change was never saved. The test then reopened the edit page and found the checkbox still checked, failing with AssertionError: assert not 'true'.

By adding the wait we make sure the redirect is there before moving on

Since `click_save()`  and  `send_invitation()`, make use of the same pattern, so this also protects the invitation flow 

Thing to mention is that if the form  fails validation, the page re-renders at the same URL instead of redirecting, so the wait raises a TimeoutException (after 10 seconds I think). 

I think that's the correct way to fail vbecause if something really is wrong with the submit, I'd rather the test blow up at the submit with a TimeoutException than mask it and surface a confusing assertion error downstream like is happening now.

### DIFF
--- a/tests/pages/pages.py
+++ b/tests/pages/pages.py
@@ -1413,6 +1413,11 @@ class InviteUserPage(BasePage):
     def send_invitation(self):
         element = self.wait_for_element(InviteUserPage.send_invitation_button)
         element.click()
+        # Wait for redirect to team members list after saving permissions or sending an invite.
+        # Without this, navigating away too quickly can abort the in-flight form POST.
+        # Note: if the form fails validation the page re-renders instead of redirecting,
+        # so this raises TimeoutException, which is the failure we'd want to see.
+        self.wait_until_url_matches(r"/services/[^/]+/users/?$")
 
     # support variants of this page with a 'Save' button instead of 'Send invitation' (both use the same locator)
     def click_save(self):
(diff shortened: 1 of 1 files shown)


risicle (approved):
Certainly good enough for a quick fix.

At some point we need to decide what part of the code should take responsibility for waiting for page change. My idea was maybe to make the `*Page` class constructors refuse to return until the url matches an expected pattern, in a way preventing the use of the class against a "wrong" page.

joybytes:
> Certainly good enough for a quick fix.
> 
> At some point we need to decide what part of the code should take responsibility for waiting for page change. My idea was maybe to make the `*Page` class constructors refuse to return until the url matches an expected pattern, in a way preventing the use of the class against a "wrong" page.

@risicle that's a great suggestion. I think I've been on firefighting mode and failed to look at the bigger picture.
I created a ticket for this as there is an upcoming epic to improve functional tests and I think this is definitely a good suggestion
https://trello.com/c/k5grM3tO/399-f

risicle:
A slightly wilder version of this idea is - make methods that change the page (click, submit etc.) responsible for _returning_ the `*Page` instance for the new page.

PR #664, opened 2026-07-30
Update  DeleteContactListPage class

kr8n3r:
Update check for ul list instead of table.
goes together with https://github.com/alphagov/notifications-admin/pull/6020

### DIFF
--- a/tests/pages/pages.py
+++ b/tests/pages/pages.py
@@ -1328,7 +1328,7 @@ class DeleteContactListPage(BasePage):
         """Returns a list of strings representing the text inside each <li>."""
         ul_element = self.wait_for_element(self.contact_list)
         items = ul_element.find_elements(*self.list_items)
-        return [item.text for item in items]
+        return [[item.text for item in items]]
 
 class CheckEmergencyContactListPage(PageWithCsvPreview):
     h1 = (By.CSS_SELECTOR, "h1")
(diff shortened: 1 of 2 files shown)


joybytes (approved):

tests/pages/pages.py
```diff
@@ -1325,11 +1324,11 @@ def click_delete(self):
...
-    def get_table_data(self):
-        table = self.wait_for_element(self.table)
-        rows = table.find_elements(*self.rows_from_table)
-        return [[cell.text for cell in row.find_elements(*self.cells_from_row)] for row in rows]
-
+    def get_list_data(self):
+        """Returns a list of strings representing the text inside each <li>."""
+        ul_element = self.wait_for_element(self.contact_list)
+        items = ul_element.find_elements(*self.list_items)
+        return [item.text for item in items]
```
joybytes: should this be `return [[item.text] for item in items]`?

joybytes:
I think on `test_upload_send_via_emergency_contact_list`
we should be calling `get_list_data` instead of `get_table_data`

tests/pages/pages.py
kr8n3r: i don't think so as this list is 1 dimensional and old table is two dimensional. i haven't been able to run this against staging, so will see if tests pass/fail
Update: yeah it needs to be as original data we assert against is in [[]]

joybytes (approved):

joybytes:
good changes, since the tests take a while to run this is not important

Currently you have `[[item.text for item in items]]`

- one recipient: `[["first@example.com"]]`
- two recipients `[["alice@example.com", "bob@example.com"]]`

if you change to `[[item.text] for item in items]`

- one recipient: `[["first@example.com"]]`
- two recipients `[["first@example.com"], ["second@example.com"]]`

Both approaches work now, because we only use one recipient, but if we have multiple it might fail. I am just not sure on the data shape to say one over the other for sure

PR #666, opened 2026-08-07
Add test_copying_an_existing_email_template

BlessedDev:
This PR adds tests covering the user journey for copying an existing templates as described in this [card](https://trello.com/c/LeJ1JIz4/717-ft-copying-templates).

### DIFF
--- a/tests/notifications/functional_tests/test_managing_templates.py
+++ b/tests/notifications/functional_tests/test_managing_templates.py
@@ -1,5 +1,7 @@
 import uuid
 
+import pytest
+
 from tests.notifications.functional_tests.test_send_files_via_ui import delete_template_from_view_email_template_page, \
     delete_template_from_view_sms_template_page, delete_template_from_view_letter_template_page
 from tests.pages import ShowTemplatesPage, ChooseExistingTemplatePage, CopyExistingTemplatePage, ViewEmailTemplatePage, \
@@ -9,6 +11,7 @@ from tests.test_utils import recordtime, go_to_templates_page, create_email_temp
 
 
 @recordtime
+@pytest.mark.xdist_group(name="copy-an-existing-template")
 def test_copying_an_existing_email_template(driver, login_seeded_user):
     go_to_templates_page(driver)
     content = "This is a test template"
@@ -45,6 +48,7 @@ def test_copying_an_existing_email_template(driver, login_seeded_user):
 
 
 @recordtime
+@pytest.mark.xdist_group(name="copy-an-existing-template")
 def test_copying_an_existing_sms_template(driver, login_seeded_user):
     go_to_templates_page(driver)
     content = "This is a test template"
@@ -81,6 +85,7 @@ def test_copying_an_existing_sms_template(driver, login_seeded_user):
 
 
 @recordtime
+@pytest.mark.xdist_group(name="copy-an-existing-template")
 def test_copying_an_existing_letter_template(driver, login_seeded_user):
     go_to_templates_page(driver)
     content = "This is a test template"
--- a/tests/notifications/functional_tests/test_managing_templates.py
+++ b/tests/notifications/functional_tests/test_managing_templates.py
@@ -1,7 +1,9 @@
 import uuid
 
-from tests.notifications.functional_tests.test_send_files_via_ui import delete_template_from_view_email_template_page
-from tests.pages import ShowTemplatesPage, ChooseExistingTemplatePage, CopyExistingTemplatePage, ViewEmailTemplatePage
+from tests.notifications.functional_tests.test_send_files_via_ui import delete_template_from_view_email_template_page, \
+    delete_template_from_view_sms_template_page
+from tests.pages import ShowTemplatesPage, ChooseExistingTemplatePage, CopyExistingTemplatePage, ViewEmailTemplatePage, \
+    ViewSMSTemplatePage
 from tests.test_utils import recordtime, go_to_templates_page, create_email_template, create_sms_template, \
     create_letter_template, delete_template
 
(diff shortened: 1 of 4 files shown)


tests/notifications/functional_tests/test_managing_templates.py
```diff
@@ -0,0 +1,64 @@
...
+    choose_existing_template_page = ChooseExistingTemplatePage(driver)
+    assert choose_existing_template_page.get_h1_text() == "Choose an existing template to copy"
+    choose_existing_template_page.click_template_by_link_text(template_name)
+    copy_existing_template_page = CopyExistingTemplatePage(driver)
+    assert copy_existing_template_page.get_h1_text() == "Copy an existing template"
+    new_template_name = f"Functional Tests - new copied {template_type} template - {uuid.uuid4()}"
+    copy_existing_template_page.change_template_name(new_template_name)
+    copy_existing_template_page.click_copy_this_template_button()
+    assert view_template_page.get_h1_text() == new_template_name
+
```
joybytes, line 54: I think to avoid flakiness we should re-instate the driver to make sure the page is correct before assertion check
```
view_template_page = view_class_name(driver)
assert view_template_page.get_h1_text() == new_template_name
```

tests/notifications/functional_tests/test_managing_templates.py
```diff
@@ -0,0 +1,64 @@
...
+    # Confirm new copy of the template was created
+    go_to_templates_page(driver)
+    assert templates_page.get_h1_text() == "Templates"
+    assert new_template_name in templates_page.get_all_listed_templates()
+
+    # Delete templates created during test run
+    templates_page.click_template_by_link_text(new_template_name)
+    delete_template_from_view_template_page(driver, new_template_name, view_template_page)
```
joybytes, line 63: same thing here just to avoid flakiness . After `delete_template_from_view_template_page` here, we should probably do

`templates_page = ShowTemplatesPage(driver)`

tests/pages/pages.py
```diff
@@ -1977,3 +2008,32 @@ def click_download_link(self):
...
+    def click_copy_this_template_button(self):
+        element = self.wait_for_element(self.copy_template_button)
+        element.click()
+
+
+def delete_template_from_view_template_page(driver, template_name, view_template_page):
+    view_template_page.click_delete_template_link()
+    view_template_page.click_template_deletion_confirmation_button()
+    view_template_page.wait_until_url_doesnt_contain("/delete")
+
```
joybytes, line 2035: maybe after waiting for the URL to not contain delete, we should give it time to fully render after redirect to avoid race conditions

```
templates_page = ShowTemplatesPage(driver)
templates_page.wait_until_url_contains("/templates")
```

tests/notifications/functional_tests/test_managing_templates.py
```diff
@@ -0,0 +1,64 @@
...
+@pytest.mark.parametrize(
+    "template_type, view_class_name",
+    [("email", ViewEmailTemplatePage), ("sms", ViewSMSTemplatePage), ("letter", ViewLetterTemplatePage)],
+)
+def test_copying_an_existing_template(driver, login_seeded_user, template_type, view_class_name):
+    go_to_templates_page(driver)
+    content = "This is a test template"
+
+    # Create a template
+    template_name = f"Functional Tests - copy email template - {uuid.uuid4()}"
```
joybytes: this runs for all template types so it should be 

`template_name = f"Functional Tests - copy {template_type} template - {uuid.uuid4()}"`

tests/notifications/functional_tests/test_managing_templates.py
```diff
@@ -0,0 +1,64 @@
...
+    create_email_template(driver, name=template_name, content=content)
+
+    # Confirm template was created
+    go_to_templates_page(driver)
+    templates_page = ShowTemplatesPage(driver)
+    assert templates_page.get_h1_text() == "Templates"
+    assert template_name in templates_page.get_all_listed_templates()
+
+    # Copy an existing template
+    view_template_page = view_class_name(driver)
```
joybytes: I think this declaration is too early, it can be removed and checked later on like I suggested

tests/notifications/functional_tests/test_managing_templates.py
BlessedDev: The page object only contains locators and the driver and not the state of pages so there is no need to re instantiate.
I deliberately use a class per page pattern, so that each page object is representative of a particular page.

tests/notifications/functional_tests/test_managing_templates.py
BlessedDev: The page object which is specifically for that page contains locators and the driver so it is not an issue when it was declared as the locators shouldn't change in this instance.

tests/notifications/functional_tests/test_managing_templates.py
BlessedDev: That is a good catch, it was oversight from the refactor. I will update it accordingly.

tests/pages/pages.py
BlessedDev: > maybe after waiting for the URL to not contain delete, we should give it time to fully render after redirect to avoid race conditions
> 
> ```
> templates_page = ShowTemplatesPage(driver)
> templates_page.wait_until_url_contains("/templates")
> ```

Yes, that is a good way to make the tests even more robust. I will make updates.

tests/notifications/functional_tests/test_managing_templates.py
BlessedDev: The page heading assertions are both for sign posting in terms of code readability and to ensure the driver is actually looking for elements as described by the locators on the right page.

tests/notifications/functional_tests/test_managing_templates.py
joybytes: I'll resolve this one as it seems like a similar comment to the ones you explained to me

thanks

